PR #223, opened 2014-12-21
Make PixelAperture.area an abstract method

cdeil:
@bsipocz @astrofrog Does it make sense to make PixelAperture.area an abstract method?

All tests pass before or after this change ... I made it because my editor was complaining here about the missing `self`.

### DIFF
--- a/photutils/aperture_core.py
+++ b/photutils/aperture_core.py
@@ -191,7 +191,8 @@ class PixelAperture(Aperture):
             if input ``error`` is not `None`.
         """
 
-    def area():
+    @abc.abstractmethod
+    def area(self):
         """
         Area of aperture.
 


larrybradley:
I think the `area()` method is useful.  It should be easy to define for any realistic (e.g. simple geometric shape) apertures.  So, I'm :+1: for making it an abstract method (with the corresponding doc changes).  I also think it should be made a "lazyproperty":
http://astropy.readthedocs.org/en/stable/api/astropy.utils.misc.lazyproperty.html?highlight=lazy#astropy.utils.misc.lazyproperty

One concrete example where `area()` is useful is local background subtraction, where one needs the mean value within the aperture.  `area()` wasn't used in the example here: http://photutils.readthedocs.org/en/latest/photutils/aperture.html#local-background-subtraction, but I will update that.

cdeil:
@larrybradley So what do I have to do ... just add `@lazyproperty` in front of the base class method?

larrybradley:
@cdeil  I've never used it with a base class.  I'd try that first, but you may need to add `@lazyproperty` in front of every `area()` method too.

cdeil:
And I guess tests should be added ... I'll try this on Saturday, but if someone gets to it before I'd be happy.

PR #240, opened 2015-01-22
Return an Astropy Table from find_peaks()

larrybradley:


cdeil:
:+1: 

Should we start a "What's new" and Changelog now?
I think both now or after the 0.2 release would be reasonable ...

larrybradley:
I started a changelog.

cdeil:
:+1: to merge.

Issue #235, opened 2015-01-15
Sub-pixel precision peak finder and image value lookup


cdeil:
@adonath and I need functions that find peaks and lookup image values with sub-pixel precision (because our images are very slow to compute at high resolution, see example [here](http://nbviewer.ipython.org/github/gammapy/gammapy-extra/blob/master/notebooks/ts_map_computation.ipynb))

@larrybradley already implemented [photutils.morphology.centroid_2dg](http://photutils.readthedocs.org/en/latest/api/photutils.morphology.centroid_2dg.html) which is one method to measure the sub-pixel peak location (using e.g. a 3 x 3 cutout image around the brightest pixel).

I think it would be nice to add a `subpixel` option [photutils.detection.find_peaks](http://photutils.readthedocs.org/en/latest/api/photutils.detection.find_peaks.html) that defaults to `False` and if set to `True` runs `centroid_2dg` on 3x3 pixel cutouts for all peaks?
Or add a separate utility function to do this?

Then we need to look up the values at these sub-pixel positions in other aligned smooth maps.
For this maybe we could add a `lookup_image_value` function that either uses `centroid_2dg` again or some interpolation method.

@larrybradley and all: Do you have any thoughts how this should be added?

larrybradley:
@cdeil  I was planning to add a feature exactly like that for `subpixel` peak finding, so I'll go ahead with that implementation.

I'm not sure exactly what you want for the look up function, but my guess is that is probably better as a separate function?

cdeil:
@larrybradley - Yes, `lookup_image_value` should be a separate function.

larrybradley:
Implemented in #244.

PR #244, opened 2015-01-28
Allow subpixel precision in find_peaks

larrybradley:
This PR implements #235.  If `subpixel=True` then the output `Table` also includes the centroid and fitted peak amplitude in addition to the (integer) peak pixel position and value.

I still need to add some tests.

This PR also removes the `segment_image` option, since one should now use `segment_properties()` when working with segmentation images.

### DIFF
--- a/photutils/detection/core.py
+++ b/photutils/detection/core.py
@@ -381,16 +381,15 @@ def find_peaks(data, threshold, box_size=3, footprint=None, mask=None,
         x_centroid, y_centroid = [], []
         fit_peak_values = []
         for (y_peak, x_peak) in zip(y_peaks, x_peaks):
-            region = cutout_footprint(data, (x_peak, y_peak),
-                                      box_size=box_size, footprint=footprint,
-                                      mask=mask, error=error)
-            gaussian_fit = fit_2dgaussian(region[0], mask=region[1],
-                                          error=region[2])
+            rdata, rmask, rerror, slc = cutout_footprint(
+                data, (x_peak, y_peak), box_size=box_size,
+                footprint=footprint, mask=mask, error=error)
+            gaussian_fit = fit_2dgaussian(rdata, mask=rmask, error=rerror)
             if gaussian_fit is None:
                 x_cen, y_cen, fit_peak_value = np.nan, np.nan, np.nan
             else:
-                x_cen = region[3][1].start + gaussian_fit.x_mean_1.value
-                y_cen = region[3][0].start + gaussian_fit.y_mean_1.value
+                x_cen = slc[1].start + gaussian_fit.x_mean_1.value
+                y_cen = slc[0].start + gaussian_fit.y_mean_1.value
                 fit_peak_value = (gaussian_fit.amplitude_0.value +
                                   gaussian_fit.amplitude_1.value)
             x_centroid.append(x_cen)
--- a/photutils/detection/core.py
+++ b/photutils/detection/core.py
@@ -253,7 +253,8 @@ def detect_sources(data, threshold, npixels, filter_kernel=None,
 
 
 def find_peaks(data, threshold, box_size=3, footprint=None, mask=None,
-               border_width=None, npeaks=np.inf, subpixel=False, wcs=None):
+               border_width=None, npeaks=np.inf, subpixel=False, error=None,
+               wcs=None):
     """
     Find local peaks in an image that are above above a specified
     threshold value.
@@ -313,6 +314,11 @@ def find_peaks(data, threshold, box_size=3, footprint=None, mask=None,
         2D Gaussian.  In this case, the fitted local centroid and peak
         amplitude will also be returned in the output table.
 
+    error : array_like, optional
+        The 2D array of the 1-sigma errors of the input ``data``.
+        ``error`` is used only to weight the 2D Gaussian fit performed
+        when ``subpixel=True``.
+
     wcs : `~astropy.wcs.WCS`
         The WCS transformation to use to convert from pixel coordinates
         to ICRS world coordinates.  If `None`, then the world
@@ -369,7 +375,8 @@ def find_peaks(data, threshold, box_size=3, footprint=None, mask=None,
         fit_peak_values = []
         for (y_peak, x_peak) in zip(y_peaks, x_peaks):
             xcen, ycen, peakval = centroid_footprint(
-                data, (x_peak, y_peak), box_size, footprint, mask)
+                data, (x_peak, y_peak), box_size=box_size,
+                footprint=footprint, mask=mask, error=error)
             x_centroid.append(xcen)
             y_centroid.append(ycen)
             fit_peak_values.append(peakval)
--- a/photutils/detection/core.py
+++ b/photutils/detection/core.py
@@ -7,7 +7,7 @@ import numpy as np
 from astropy.table import Column, Table
 from astropy.convolution import Kernel2D
 from astropy.stats import sigma_clipped_stats
-from ..morphology import centroid_footprint
+from ..morphology import cutout_footprint, fit_2dgaussian
 from ..utils.wcs_helpers import pixel_to_icrs_coords
 
 
@@ -273,9 +273,9 @@ def find_peaks(data, threshold, box_size=3, footprint=None, mask=None,
 
     When using subpixel precision (``subpixel=True``), then a cutout of
     the specified ``box_size`` or ``footprint`` will be taken centered
-    on each peak and fit with a 2D Gaussian.  In this case, the fitted
-    local centroid and peak amplitude will also be returned in the
-    output table.
+    on each peak and fit with a 2D Gaussian (plus a constant).  In this
+    case, the fitted local centroid and peak amplitude will also be
+    returned in the output table.
 
     Parameters
     ----------
@@ -285,15 +285,20 @@ def find_peaks(data, threshold, box_size=3, footprint=None, mask=None,
     threshold : float
         The data value to be used for the detection threshold.
 
-    box_size : int, optional
-        The square box size (pixel) of the local region to search for
-        peaks at every point in ``data``.
+    box_size : scalar or tuple, optional
+        The size of the local region to search for peaks at every point
+        in ``data``.  If ``box_size`` is a scalar, then the region shape
+        will be ``(box_size, box_size)``.  Either ``box_size`` or
+        ``footprint`` must be defined.  If they are both defined, then
+        ``footprint`` overrides ``box_size``.
 
     footprint : `~numpy.ndarray` of bools, optional
         A boolean array where `True` values describe the local footprint
         region within which to search for peaks at every point in
-        ``data``.  ``footprint`` overrides ``box_size`` when defining
-        the local region around each pixel.
+        ``data``.  ``box_size=(n, m)`` is equivalent to
+        ``footprint=np.ones((n, m))``.  Either ``box_size`` or
+        ``footprint`` must be defined.  If they are both defined, then
+        ``footprint`` overrides ``box_size``.
 
     mask : array_like, bool, optional
         A boolean mask with the same shape as ``data``, where a `True`
@@ -311,8 +316,9 @@ def find_peaks(data, threshold, box_size=3, footprint=None, mask=None,
     subpixel : bool, optional
         If `True`, then a cutout of the specified ``box_size`` or
         ``footprint`` will be taken centered on each peak and fit with a
-        2D Gaussian.  In this case, the fitted local centroid and peak
-        amplitude will also be returned in the output table.
+        2D Gaussian (plus a constant).  In this case, the fitted local
+        centroid and peak amplitude will also be returned in the output
+        table.
 
     error : array_like, optional
         The 2D array of the 1-sigma errors of the input ``data``.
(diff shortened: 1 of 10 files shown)


cdeil:
@larrybradley Thanks for implementing this!

@adonath and I just discussed this a bit (but didn't try or test it yet) ... here's some comments:
- This will not fail with thresholds and peaks below zero, no? It should just be a few line changes to support this.
- Does `fit_gauss2d` always work or will can it fail for pathological cases like plateaus or random noise input? If it can throw exceptions or raise warnings maybe these should be caught and the corresponding values in the output table set to `nan` or some other mechanism that allows getting the results for all the other peaks.
- Do you think it would be worth adding a convenience where additionally world coordinates are computed if the user passes a `wcs`? Or course users can do this after running this function, but it's a few lines of code and a very common use case, so we think it might be nice to add this.

larrybradley:
Negative thresholds and peaks should work fine.  It's simply an `data > theshold` test.

I'm sure `fits_gauss2d` can fail in some cases.  In some cases (e.g. a single pixel), it will raise a warning (`WARNING: The fit may be unsuccessful; check fit_info['message'] for more information. [astropy.modeling.fitting]`), but the resulting fit is actually OK.  Unfortunately I suspect that a complete failure also raises the same warning (some fits with warnings are probably OK).  I'll try to catch the warnings and return `nan`s when it occurs.

I can add the `wcs` convenience.  Can you suggest names for these columns (e.g., `ra_peak_icrs`, `dec_peak_icrs`, `fit_ra_peak_icrs`, `fit_dec_peak_icrs`)?

photutils/detection/core.py
```diff
@@ -295,39 +302,79 @@ def find_peaks(data, threshold, min_separation=2, exclude_border=True,
...
+
+    if footprint is not None:
+        data_max = ndimage.maximum_filter(data, footprint=footprint,
+                                          mode='constant', cval=0.0)
+    else:
+        data_max = ndimage.maximum_filter(data, size=box_size,
+                                          mode='constant', cval=0.0)
+
+    peak_data = data.copy()
+    peak_data *= (data == data_max)
```
cdeil: Here you zero out pixels that are not peaks:
```
peak_data *= (data == data_max)
```
What if `theshold=-42`, won't these regions become peaks?

photutils/detection/core.py
```diff
@@ -295,39 +302,79 @@ def find_peaks(data, threshold, min_separation=2, exclude_border=True,
...
+
+    if border_width is not None:
+        for i in range(peak_data.ndim):
+            peak_data = peak_data.swapaxes(0, i)
+            peak_data[:border_width] = 0.
+            peak_data[-border_width:] = 0.
```
cdeil: Here you use `0` as a special value ... what if my peak pixel has value 0 because all the rest of the image is negative?

As I said, I didn't test, but it looks to me this might not work in that case?

cdeil:
Concerning which world coordinates to put and what their names should be, my suggestion would be to do whatever SExtractor does. But I think we had this discussion already in the context of [aperture_photometry](http://photutils.readthedocs.org/en/latest/api/photutils.aperture_photometry.html#photutils.aperture_photometry) or somewhere else and did something different ... I don't remember.

cdeil:
Obviously only call it RA / DEC and ICRS if you decide to always do a celestial coordinate transform to that system ... I think just giving the plain world coordinates as columns with unit `radian` or `deg` and letting the user figure out his celestial system would be fine, too.

photutils/detection/core.py
larrybradley: Negative values in `peak_data` should be OK.  The peaks are identified using `.nonzero()`.  But as you point out in the next note, a `peak value == 0` is problematic.  I can fix that.

photutils/detection/core.py
larrybradley: Yes, a peak value of 0 is a problem...

photutils/detection/core.py
larrybradley: I take that back!  Yes, this too will fail and I need to change some things for non-positive thresholds.

larrybradley:
@cdeil Zero or negative peaks now work for the integer peak finder.  However, I ran into an issue with the subpixel finder using a 2D Gaussian fit.  It really requires a 2D compound model (constant + Gaussian2D).  With `astropy 0.4`, that would require creating a custom model (compound 2D models weren't possible).  I think 2D compound models work now in `astropy 1.0`.  So, I could either create a custom compound model, or perhaps better is that we drop 0.4 support.

Should we consider a `photutils 0.2` release that requires `astropy 1.0` soon after that release?  We could also remove *all* of the stuff that has been building up in `photutils.extern`.

cdeil:
I'm +1 to drop Astropy 0.4 support ... moving the discussion to a separate issue: #246.

Concerning the peak fitting, I can see how adding a constant would help a lot, especially if there's many more pixels than free parameters participating in the fit.

I haven't looked into sub-pixel peak location methods before, but a quick Google search suggests a 2D parabola might also yield good results (see [here](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.380.5886&rep=rep1&type=pdf) or [here](http://www-ist.massey.ac.nz/dbailey/sprg/pdfs/2003_IVCNZ_414.pdf)).
@larrybradley Do you think Const + Gaussian works for all the applications you have (I think it does for Gammapy) or is it worth setting this up so that it can be extended to apply a few select common methods?

adonath:
@larrybradley Thanks for working on this!

Currently `find_peaks` fails, when there are `NaN` values present in the peak region. Here is an example, that raises a `LinAlgError`. 
```
data = np.zeros((5, 5))
data[1:4, 1:4] = 1
data[2, 2] = 2
data[3, 2] = np.nan
find_peaks(data, 1, subpixel=True, footprint=np.ones((3,3)))
```
I know that this is a rare case, but wouldn't it make sense to have an additional `mask` argument in the `find_peaks` function to mask out `NaN` values or regions, where you don't want to find peaks (e.g. because they are very noisy...)? As `fit_2dgaussian` also takes a mask, `NaN` values could be just excluded from the fit.

Concerning the problem with the Gaussian fit: I don't know whether this works, but one could subtract the minimum in the region, from the values that are fitted. Or use the mean/median value of the pixels in a 'ring' around the region that is fitted, as a local estimate of the background and subtract it.

larrybradley:
@adonth  Thanks for those suggestions!  I've added the `mask` option and I'm also now subtracting the local minimum as a background estimate.  That seems to work fine and is a lot of simpler.  Given small footprint sizes, I would guess that a 2D Gaussian + constant would give very similar results.

photutils/detection/core.py
```diff
@@ -250,19 +254,29 @@ def detect_sources(data, threshold, npixels, filter_kernel=None,
...
+    Peaks are the maxima above the ``threshold`` within a local region.
+    The regions are defined by either the ``box_size`` or ``footprint``
+    parameters.  ``box_size`` defines the local region around each pixel
+    as a square box.  ``footprint`` is a boolean array where `True`
+    values specify the region shape.
 
-    If peaks are flat (i.e., multiple adjacent pixels have identical
-    intensities), then the coordinates of all such pixels are returned,
-    even if they are not separated by at least ``min_separation``.
+    If mulitple pixels within a local region have identical intensities,
```
cdeil: Typo: `mulitple` -> `multiple`

photutils/detection/core.py
```diff
@@ -250,19 +254,29 @@ def detect_sources(data, threshold, npixels, filter_kernel=None,
...
+    then the coordinates of all such pixels are returned.  Otherwise,
+    there will be only one peak pixel per local region.  Thus, the
+    defined region effectively imposes a minimum separation between
+    peaks (unless there are identical peaks within the region).
+
+    When using subpixel precision (``subpixel=True``), then a cutout of
+    the specified ``box_size`` or ``footprint`` will be taken centered
+    on each peak and fit with a 2D Gaussian.  In this case, the fitted
```
cdeil: This sentence has to be updated with whatever exact method you choose: currently "subtract const, then fit 2d gauss".

larrybradley:
@cdeil I saw some unexpectedly large shifts (~0.2 pixels) when writing some tests for this PR, so I'm actually working on adding the Const+2DGaussian now.

larrybradley:
I've changed this PR to fit with a Const+Gaussian2D, which gives more robust results.  This PR should be ready to go now.

photutils/detection/core.py
```diff
@@ -295,39 +313,112 @@ def find_peaks(data, threshold, min_separation=2, exclude_border=True,
...
+        coordinates will not be returned in the output
+        `~astropy.table.Table`.
 
     Returns
     -------
     output : `~astropy.table.Table`
         A table containing the x and y pixel location of the peaks and
-        their values.
+        their values.  If ``subpixel=True``, then the table will also
+        contain the local centroid and fitted peak value.
```
cdeil: From the code it looks like the peak value you return is relative to the background level (which is not obtainable by the user). I'm not sure what's best to do here ... return background level as well or not return the peak value, or just be more explicit in the description here?

larrybradley: It actually returns the Gaussian2D amplitude + constant value.  It's value should be close to the peak pixel value.  I'll update the docs here to make that clearer.

photutils/utils/wcs_helpers.py
```diff
@@ -90,3 +90,39 @@ def assert_angle(name, q):
...
+        to ICRS world coordinates.
+        `~astropy.table.Table`.
+
+    Returns
+    -------
+    ra : `~astropy.units.Quantity`
+        The ICRS Right Ascension in degrees.
+
+    dec : `~astropy.units.Quantity`
+        The ICRS Right Ascension in degrees.
```
cdeil: `Right Ascension` -> `Declination`

photutils/detection/core.py
cdeil: Ah, Ok ... then I misread the code ... that's actually what the docs say already ... maybe add "absolute" peak value to be super-clear.

photutils/utils/wcs_helpers.py
larrybradley: Done, thanks :/

photutils/detection/core.py
```diff
@@ -295,39 +314,112 @@ def find_peaks(data, threshold, min_separation=2, exclude_border=True,
...
-    return Table((x_peaks, y_peaks, peak_values),
-                 names=('x_peak', 'y_peak', 'peak_value'))
+    if subpixel:
+        x_centroid, y_centroid = [], []
+        fit_peak_values = []
+        for (y_peak, x_peak) in zip(y_peaks, x_peaks):
+            region = cutout_footprint(data, (x_peak, y_peak),
+                                      box_size=box_size, footprint=footprint,
+                                      mask=mask, error=error)
+            gaussian_fit = fit_2dgaussian(region[0], mask=region[1],
```
cdeil: IMO if you return more than one or two things it's better to return a bunch or dict or `namedtuple`, i.e. give the things you return names.
This makes for more readable code when you use some of the returned results ... in this case I have to look at the docs to be sure what [0], [1] and [2] are.
But I know that others prefer to just return long tuples ... so if you prefer this, OK!

cdeil:
@larrybradley - Looks great ... can't wait to use this ... thanks!

photutils/morphology.py
```diff
@@ -5,14 +5,27 @@
...
+from astropy.modeling.models import Gaussian1D, Gaussian2D, Const1D, Const2D
 from astropy.modeling.fitting import LevMarLSQFitter
+from astropy.nddata.utils import overlap_slices
 from .segmentation import SegmentProperties
+import warnings
 
 
-__all__ = ['centroid_com', 'gaussian1d_moments', 'centroid_1dg',
-           'centroid_2dg', 'fit_2dgaussian', 'data_properties']
+__all__ = ['GaussianConst1D', 'GaussianConst2D', 'centroid_com',
```
cdeil: Should we expose `GaussianConst1D` and `GaussianConst2D` in the docs (i.e. kind of make them part of our public API)?
The main issue with this I see is that the docstring doesn't even mention the parameter names (or does it via some Python trickery?), i.e. makes this model hard to use.

larrybradley: I'm going to take them out of the public API.  There's also another issue with the docs for these compound models.  They get base classes called `photutils.morphology.CompoundModel0` and
`photutils.morphology.CompoundModel1`, which do not explicitly exist and therefore do not resolve (giving `Sphinx` warnings):
untitled

@embray:  can the `Sphinx` warnings for compound models be fixed?

cdeil: Now it gets interesting ... we need lots of compound models in Gammapy ... is it possible to make them well-documented or is the simplest way to just sub-class `Model` and define it from scratch by calling the component `evaluate` static methods in the compound model `evaluate` method?

larrybradley: I don't know.  @embray?

embray: In general, with the `..autoclass::` directive, you can add `:no-show-inheritance` to the directive to disable that.  Though I'm not sure how to work that in the context of the module autosummary.

Another thing you can do is add to nitpick_ignore in conf.py a la

https://github.com/astropy/astropy/blob/master/docs/conf.py#L193 and
https://github.com/astropy/astropy/blob/master/docs/nitpick-exceptions

I wonder if that supports patterns though.  Because really you want to ignore something like `photutils.morphology.CompoundModel*`.

A final possiblity is that I might might be able to just totally munge the `__bases__` for subclasses of generated compound models like the ones in this module.  I'll have to have a look at that sometime.

larrybradley: @cdeil  I kept `Gaussian2DConst` in the public API because that's what `fit_2dgaussian` returns.  I  added its parameter documentation.

larrybradley: Thanks, @embray.  `:no-show-inheritance` didn't work, so I added a nitpick exception (I didn't try a pattern).

photutils/detection/core.py
larrybradley: I changed this to return a tuple.  Thanks.

Issue #46, opened 2014-05-16
Introduce PixelCoordinates class


cdeil:
As discussed in today's photutils hangout (see [minutes](https://docs.google.com/document/d/1j6L0XhQFVQpB8Bc1BQ3j813_mJdjDP4dkbkfgTYOBfY)), we propose to introduce a `PixCoord` class and use it throughout `photutils`.

This is a first step towards a simple and consistent API. Currently e.g. the [aperture_photometry](http://photutils.readthedocs.org/en/latest/api/photutils.aperture_photometry.html)  function takes two arguments `xc` and `yc` and the `psf_photometry` function (to be merged today) takes a single argument `positions`. Soon we also want to support [SkyCoord](http://docs.astropy.org/en/latest/api/astropy.coordinates.SkyCoord.html) coordinates, at least in some convenience functions ... this will be discussed in a separate issue later.

Once [Generalised WCS](https://github.com/astropy/astropy-api/pull/10) becomes available in Astropy it will probably contain a class similar to `PixCoord` and we'll use it, but for now we'll just make our own in `photutils`.

cc @astrofrog @eteq @bsipocz

PR #261, opened 2015-04-07
Change centroid functions output to be in (y, x) order

larrybradley:
This makes the centroid functions consistent with `SegmentProperties` centroids.

### DIFF
--- a/docs/photutils/morphology.rst
+++ b/docs/photutils/morphology.rst
@@ -33,21 +33,21 @@ subtract the background)::
 
 .. doctest-requires:: skimage
 
-    >>> x1, y1 = centroid_com(data)
-    >>> print(x1, y1)    # doctest: +FLOAT_CMP
-    (13.93157998341213, 17.051234441067088)
+    >>> y1, x1 = centroid_com(data)
+    >>> print(y1, x1)    # doctest: +FLOAT_CMP
+    (17.051234441067088, 13.93157998341213)
 
 .. doctest-requires:: scipy
 
-    >>> x2, y2 = centroid_1dg(data)
-    >>> print(x2, y2)    # doctest: +FLOAT_CMP
-    (14.040352707371396, 16.962306463644801)
+    >>> y2, x2 = centroid_1dg(data)
+    >>> print(y2, x2)    # doctest: +FLOAT_CMP
+    (16.962306463644801, 14.040352707371396)
 
 .. doctest-requires:: scipy, skimage
 
-    >>> x3, y3 = centroid_2dg(data)
-    >>> print(x3, y3)    # doctest: +FLOAT_CMP
-    (14.002212073733611, 16.996134592982017)
+    >>> y3, x3 = centroid_2dg(data)
+    >>> print(y3, x3)    # doctest: +FLOAT_CMP
+    (16.996134592982017, 14.002212073733611)
 
 Now let's plot the results.  Because the centroids are all very
 similar, we also include an inset plot zoomed in near the centroid:
@@ -86,9 +86,9 @@ similar, we also include an inset plot zoomed in near the centroid:
                                       centroid_2dg)
     import matplotlib.pyplot as plt
     data = make_4gaussians_image()[43:79, 76:104]    # extract single object
-    x1, y1 = centroid_com(data)
-    x2, y2 = centroid_1dg(data)
-    x3, y3 = centroid_2dg(data)
+    y1, x1 = centroid_com(data)
+    y2, x2 = centroid_1dg(data)
+    y3, x3 = centroid_2dg(data)
     fig, ax = plt.subplots(1, 1)
     ax.imshow(data, origin='lower', cmap='Greys_r')
     marker = '+'
--- a/photutils/morphology.py
+++ b/photutils/morphology.py
@@ -114,16 +114,16 @@ def centroid_com(data, mask=None):
 
     Returns
     -------
-    xcen, ycen : float
-        (x, y) coordinates of the centroid.
+    ycen, xcen : float
+        (y, x) coordinates of the centroid.
     """
 
     from skimage.measure import moments
     data = _convert_image(data, mask=mask)
     m = moments(data, 1)
-    xcen = m[1, 0] / m[0, 0]
     ycen = m[0, 1] / m[0, 0]
-    return xcen, ycen
+    xcen = m[1, 0] / m[0, 0]
+    return ycen, xcen
 
 
 def gaussian1d_moments(data, mask=None):
--- a/photutils/morphology.py
+++ b/photutils/morphology.py
@@ -189,17 +189,17 @@ def marginalize_data2d(data, error=None, mask=None):
 
     if error is not None:
         marginal_error = np.array(
-            [np.sqrt(np.sum(error**2, axis=i)) for i in [0, 1]])
+            [np.sqrt(np.sum(error**2, axis=i)) for i in [1, 0]])
     else:
         marginal_error = [None, None]
 
     if mask is not None:
         mask = np.asanyarray(mask)
-        marginal_mask = [np.sum(mask, axis=i).astype(np.bool) for i in [0, 1]]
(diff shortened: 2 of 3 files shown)


cdeil:
Is `(y, x)` the way to go here?

I assumed that for the end-user interface we use `(x, y)` order where we use pixel coordinate tuples and can't avoid the ambiguity by using `Table` objects.
The aperture classes (see e.g. [CircularAperture](http://photutils.readthedocs.org/en/latest/api/photutils.aperture_core.CircularAperture.html)) use `positions=(x, y)` tuples as input).

I know `(y, x)` has the advantage that one can directly use it to index into numpy arrays, so I'm not saying I'm strongly pro `(x, y)`, but I think it should be carefully considered which is better.

Maybe at least always using [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) and encouraging users to write code that indexes `x` and `y` by name instead of by index would help?

larrybradley:
I like `(y, x)` better because it can be used directly for array indexing.  In principle, one could argue that the aperture `positions` should also be in `(y, x)` axis order.  It would be nice to have a consistent API that would also work with array indexing, but perhaps changing everything to `(y, x)` will cause too much confusion?

Using a `namedtuple` has the advantage of making `x` and `y` explicit, which is nice, at the expense of making direct array indexing a little more cumbersome.  Even so, that's probably better than having to flip an `(x, y)` index (or remember that you *need* to flip `x` and `y`).

Does anyone else have other thoughts before I change this PR to use `namedtuple`s?

cdeil:
> It would be nice to have a consistent API that would also work with array indexing, but perhaps changing everything to (y, x) will cause too much confusion?

I think a consistent API for photutils 0.2 is **much** more important that backwards compatibility with photutils 0.1 ... the joys of the 1000s of future users versus the pain of the 10s of current users ... :-)

> Using a namedtuple has the advantage of making x and y explicit, which is nice, at the expense of making direct array indexing a little more cumbersome.

@larrybradley – How is it more cumbersome? I haven't used `namedtuple` much, but this seems to work OK:
```python
>>> import numpy as np
>>> from collections import namedtuple
>>> PixPosition = namedtuple('PixPosition', ['y', 'x'])
>>> data = np.arange(15).reshape((5, 3))
>>> pos1 = PixPosition(y=1, x=2)
>>> pos2 = (1, 2)
>>> print(pos1)
PixPosition(y=1, x=2)
>>> print(data[pos1])
5
>>> print(pos2)
(1, 2)
>>> print(data[pos2])
5
```

I now remembered #46 which is related to whether to use `namedtuple` here or not.
PS: I'm not the only one that thinks a pixel position class and using it as much as possible in the end-user API to avoid the possibility of user error is a good idea (see [here](http://thetractor.org/doc/intro.html#sources-positions-and-brightnesses)).

larrybradley:
@cdeil  I was thinking that one would have to use `data[pos1.y, pos1.x]`.  But I see that if the `namedtuple` is defined with `'y', 'x'` order then all is fine with `data[pos]`!  I don't see any downside to `namedtuples` now, so I'm going to use them.  Thanks!

Any thoughts for the aperture `positions` inputs?

cdeil:
@astrofrog , @bsipocz This is a fundamental decision for photutils, please comment!
(looking back at #46 the minutes from May 2014 say a pixel coordinate class should be introduced, and then there's a comment from August 2014 without a reference saying this proposal has been rejected)

> Any thoughts for the aperture positions inputs?

I've thought about this some more and I'm pretty sure I'm  +1 for `x, y` order. My main argument is that in astronomy images usually `x` is roughly aligned with `longitude` and `y` with `latitude` and `astropy.wcs` as well as `astropy.coordinates` (see e.g. [SkyCoord.from_pixel](http://astropy.readthedocs.org/en/latest/api/astropy.coordinates.SkyCoord.html#astropy.coordinates.SkyCoord.from_pixel)) use `x, y` order.

As far as I can see the main use of `y, x` order is in the [nddata.utils](http://astropy.readthedocs.org/en/latest/nddata/index.html#id1) functions, which makes sense because they are mainly used for indexing and slicing numpy arrays.

Using a class instead of a namedtuple would allow adding an attribute to easily index into numpy arrays without the user even having to remember an order  (the `PixCoord` class knows what order Numpy expects and stores `x` and `y` internally, no tuple, no problems). Here's what it could look like:
```python
import numpy as np
from astropy.coordinates import Angle, PixCoord, SkyCoord
from photutils import CircularAperture, SkyCircularAperture
pix_aperture = CircularAperture(positions=PixCoord(x=42, y=43), radius=3)
sky_aperture = SkyCircularAperture(positions=SkyCoord(42, 43, unit='deg'), r=Angle(3, 'deg'))
pix_pos = PixCoord(x=1, y=2)
data = np.arange(15).reshape((5, 3))
data[pix_pos.numpy_index]
```

astrofrog:
I need to think about it. At first glance I personally think we should *always* use x, y as the order, except for indices. This is what is used in most other places. For example the WCS class takes (x, y). I do like the possibility of using some kind of class/namedtuple to make this more straightforward for users, but am just (at the moment) opposed to still returning two arrays and just swapping them.

Another initial reaction is that I think there are more current users than you might think, and if we silently change the meaning of x, y -> y, x that could cause some majorly wrong results (not just an inconvenience). Even at this stage, I would say that we should preserve backward compatibility unless there is no other option than to break it and there is a clear benefit (not just 'elegance' for us). I've heard users who were pretty annoyed that things in Astropy 0.2 no longer worked - it doesn't matter if we tell users it's experimental or not, the more we break, the more we annoy them, the less they are likely to recommend the package to others. We will only reach 1000s of users if the early adopters like it and recommend it! So let's break the API only if there is a compelling reason, and there's actually no way to maintain back-compatibility.

larrybradley:
I am closing the PR and leaving things in `(x, y)` order.  To make everything consistent, unfortunately I'll need to change the tuple order in a few methods of `segmentation.SegmentProperties` (e.g. `centroid`, `local_centroid`, etc.).   I'll do that in a separate PR.  Those specific properties probably aren't used much directly, so fortunately it should not have much impact (the table returned by `properties_table()` has separate `x` and `y` columns).

Should the `PixCoord` class go in the `astropy` core?

rlpitts:
This issue needs revisiting. I tried working exclusively in sky coordinates specifically to avoid the (x,y) vs (y,x) order confusion, and my SkyEllipticalAperture instance still landed in the wrong place. 
Let core = a SkyCoords instance, hdr = a fits header, and w = the corresponding WCS. If I do this,
`>>> aptr = photutils.SkyEllipticalAperture(core, hdr['BMAJ']*units.arcsec,
                                             hdr['BMIN']*units.arcsec,
                                             theta=hdr['BPA']*units.deg)`
`>>> phot_table = photutils.aperture_photometry(hdu[0].data, aptr,wcs=w)`
or replace that last line with 
`>>> phot_table = photutils.aperture_photometry(hdu[0].data, aptr.to_pixel(w))`
the aperture sum I get is vastly different than the correct sum (checked with imshow) I get when I give up and do everything in pixel coordinates:
`>>> y,x=np.squeeze(cygcd['S8'].to_pixel(wcs.WCS(hdu[0].header)))`
`>>> aptr = photutils.EllipticalAperture([y,x],hdr['BMAJ']/hdr['CDELT2'],
                                             hdr['BMIN']/hdr['CDELT2'],
                                             theta=(hdr['BPA']+90.)*np.pi/180.)`
`>>> phot_table = photutils.aperture_photometry(hdu[0].data, aptr, 
                                                    error=uhdu[0].data)`

larrybradley:
@rlpitts The `SkyCoord.to_pixel()` method returns `(x, y)` not `(y, x)` as you have above `y,x=np.squeeze(core].to_pixel(wcs.WCS(hdu[0].header)))`.

There isn't any (x, y) confusion in `photutils`.  The aperture positions are either `(x, y)` or a list of `(x, y)` pairs.  Under the hood, `photutils` is simply using `astropy.wcs` methods for conversion between pixel and sky (and visa versa).  See details here:  https://docs.astropy.org/en/stable/wcs/wcsapi.html.

To show code blocks you can surround the code in triple backticks.

PR #269, opened 2015-05-22
Update sigma clipping based on updated astropy

larrybradley:
Tests are currently failing because of changes in `sigma_clip` and `sigma_clipped_stats` in astropy master.  There is also a small bug in astropy's `sigma_clip` that I need to fix where a scalar mask is returned when no values are clipped.  In that case `ma.data[~ma.mask]` doesn't give the desired result.

### DIFF
--- a/photutils/background.py
+++ b/photutils/background.py
@@ -7,9 +7,11 @@ from astropy.stats import sigma_clip
 from astropy.utils import lazyproperty
 import warnings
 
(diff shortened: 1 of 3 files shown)


larrybradley:
The tests will pass with this PR once the upstream fixes in `astropy` (https://github.com/astropy/astropy/pull/3792) are merged.

PR #286, opened 2015-10-08
Update astropy-helpers to 1.0.5 (and fix failing tests)

larrybradley:
The failing doc tests in https://github.com/astropy/photutils/pull/284 and https://github.com/astropy/photutils/issues/285 are not due to the `astropy-helpers` update.  They occurred because the `sigma_clipped_stats` `iters` default value changed from `None` to `5` in `astropy dev`.  In this PR, I explicitly set `iters=5` so all versions of `astropy` are happy.

### DIFF
--- a/ah_bootstrap.py
+++ b/ah_bootstrap.py
@@ -409,7 +409,7 @@ class _Bootstrapper(object):
     def get_index_dist(self):
         if not self.download:
             log.warn('Downloading {0!r} disabled.'.format(DIST_NAME))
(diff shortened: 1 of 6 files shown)


larrybradley:
I tracked down the new failing tests.  Due to a change in `numpy 1.10`, `assert_allclose` no longer works with `nddata` objects with units.  I've updated the tests to properly use `astropy.tests.helper.assert_quantity_allclose`.

PR #310, opened 2015-12-26
generalized interpolate_masked_data to n dimensions

larsbratholm:
tested for 2-7 dimensions


larrybradley:
Thanks, @larsbratholm!  It would be good to have some unit tests for the higher dimensions.  Could you add some (to `test_interpolation.py`)?

larrybradley:
Aside from some minor PEP8 nitpicks (which I'll fix), this looks good.  Thanks again, @larsbratholm.

PR #319, opened 2016-01-13
Fix matplotlib version to <=1.5.0 using pip for travis-ci sphinx builds

larrybradley:


larrybradley:
This needs be fixed upstream in `ci-helpers` because it automatically installs `matplotlib` for `build_sphinx`.

PR #321, opened 2016-01-28
find_peaks works well with a threshold array

hamogu:
I originally planned to implement a way where the detection threshold could be an array for `find_peaks` (which sounds like a simple extension of the existing algorithm, since find_peaks works local anyway), but then I found that that works already.
Thus, I now only propose this minor update to the docs.

[skip ci]


larrybradley:
Thanks, @hamogu!

PR #324, opened 2016-02-05
Retrieve the fractions associated with a given aperture/photometry

fdeugenio:
At the moment the methods do_photometry & co. accept only 2d input data.

This is problematic when doing aperture photometry on a data cube (e.g. create aperture spectra).
The fastest method is to get the pixel weights (fractions) directly.

An alternative implementation would see heavier edits, with an added ``axes`` keyword which specifies which axes will be suppressed by ``do_photometry``.

### DIFF
--- a/photutils/aperture_core.py
+++ b/photutils/aperture_core.py
@@ -18,8 +18,9 @@ from astropy.utils.misc import InheritDocstrings
 from astropy.utils.exceptions import AstropyUserWarning
 from astropy.wcs.utils import skycoord_to_pixel
 from astropy.nddata import support_nddata
-from .aperture_funcs import (do_circular_photometry, do_elliptical_photometry,
-                             do_rectangular_photometry)
+from .aperture_funcs import (do_circular_photometry, get_circular_fractions,
+                             do_elliptical_photometry, get_elliptical_fractions,
+                             do_rectangular_photometry, get_rectangular_fractions)
 from .utils.wcs_helpers import (skycoord_to_pixel_scale_angle, assert_angle,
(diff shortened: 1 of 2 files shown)


photutils/aperture_core.py
```diff
@@ -186,7 +187,47 @@ def do_photometry(self, data, error=None, effective_gain=None,
...
+        subpixels : int, optional
+            For the ``'subpixel'`` method, resample pixels by this factor (in
+            each dimension). That is, each pixel is divided into
+            ``subpixels ** 2`` subpixels.
+
+        Returns
+        -------
+        fraction : `numpy.array`
+            Array with the same shape as ``data``. Each element is the fraction
+        of the corresponding ``data`` pixel that falls within the aperture.
```
larrybradley: This line should be indented to match the line above.

larrybradley:
Thanks, @fdeugenio!  This looks good.  Could you please add some tests for the new methods?

larrybradley:
And it would be great to perhaps include a simple test for your use case of performing aperture photometry on a (small) data cube.

fdeugenio:
dear @larrybradley , I can certainly add some tests - but the timescale for this would be about 1.5 months. Is that still ok?

larrybradley:
In that case I'll probably merge this soon.  You can add the tests later if they haven't already been added.  Thanks.

fdeugenio:
Sorry Larry,

I am not allowed can't take my laptop on holidays. I'll see what can I 
do from my phone though :-(

larrybradley:
No worries.  Thanks for your contribution, @fdeugenio.

PR #335, opened 2016-03-14
Skeleton structure for Polygon Overlap Grid

cheekujodhpur:
For having a Polygon aperture as in #131 , I am starting with a Triangular Aperture.

The attached code is the skeleton for calculating triangular overlap grid.

### DIFF
--- a/photutils/geometry/__init__.py
+++ b/photutils/geometry/__init__.py
@@ -6,7 +6,9 @@ Geometry subpackage for low-level geometry functions.
 from .circular_overlap import *
 from .elliptical_overlap import *
 from .rectangular_overlap import *
+from .polygon_overlap import *
 
 __all__ = ['circular_overlap_grid',
            'elliptical_overlap_grid',
-           'rectangular_overlap_grid']
+           'rectangular_overlap_grid',
+           'polygon_overlap_grid']
--- a/photutils/geometry/polygon_overlap.pyx
+++ b/photutils/geometry/polygon_overlap.pyx
@@ -1,0 +1,124 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+import numpy as np
+cimport numpy as np
+
+__all__ = ['polygon_overlap_grid']
+
+DTYPE = np.float64
+ctypedef np.float64_t DTYPE_t
+
+cimport cython
+
+def polygon_overlap_grid(double xmin, double xmax, double ymin,
+                            double ymax, int nx, int ny, vertices,
+                            int use_exact, int subpixels):
+    """
+    polygon_overlap_grid(xmin, xmax, ymin, ymax, nx, ny,
+                            vertices, use_exact, subpixels)
+
+    Area of overlap between a polygon and a pixel grid.
+
+    Parameters
+    ----------
+    xmin, xmax, ymin, ymax : float
+        Extent of the grid in the x and y direction
+    nx, ny : int
+        Grid dimensions
+    vertices : list of tuples of length 2
+        Vertex co-ordinates for the polygon
+    theta : float
+        The position angle of the side represented by ``a``
+        in radians (counterclockwise).
+    use_exact : 0 or 1
+        If set to 1, calculates the exact overlap, while if set to 0, uses a
+        subpixel sampling method with ``subpixel`` subpixels in each direction
+    subpixels : int
+        If ``use_exact`` is 0, each pixel is resmapled by this factor in each dimension. Thus, each pixel is divided into ``subpixels ** 2`` subpixels.
+
+    Returns
+    -------
+    frac : `~numpy.ndarray`
+        2-d array giving the fraction of the overlap
+    """
+
+    cdef unsigned int i, j 
+    cdef double x, y, dx, dy
+    cdef double pxmin, pxmax, pymin, pymax
+
+    # Define output array
+    cdef np.ndarray[DTYPE_t, ndim=2] frac = np.zeros([ny,nx],dtype=DTYPE)
+
+    if use_exact == 1:
+        raise NotImplementedError("Exact mode has not been implemented for "
+                                  "rectangular apertures")
+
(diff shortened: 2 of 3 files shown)


photutils/geometry/triangular_overlap.pyx
```diff
@@ -0,0 +1,101 @@
...
+
+    Parameters
+    ---------
+    xmin, xmax, ymin, ymax : float
+        Extent of the grid in the x and y direction
+    nx, ny : int
+        Grid dimensions
+    a, b, c : float
+        Sides of the triangle
+    theta : float
```
hamogu: For a triangle, would it not make more sense to allow a free specification of the three corners?
I did not quite follow the individual formulas, but if I can specify the triangle with a,b,c, and theta it must be some special case of a triangle (maybe is one angle assume to be 90 deg?)
At the very least, this should be called "EqualsidedTriangle" or "right-angles triangle" (or whatever the assumption is).

cheekujodhpur: I did start with allowing a free specification of the three corners, but I later realized that specifying an aperture just involves the shape and size of the aperture and the center of it is assumed to be the center of the extensions on the image, as in the case of rectangular aperture.

The three side lengths uniquely specify a triangle. Theta is used to specify the orientation of this triangle in space.

Does that make more sense?

hamogu:
I've never used Cython myself, so I can't really say much about the structure of the code here.

This PR still needs tests, documentation and an entry in the CHANGES.rst file.
When the test is written it will become clearer that the code works, too even for me.

cheekujodhpur:
Completed the function and wrote tests.

This is a crucial function to implement triangular apertures, but has no direct use so I do not feel a need of writing a user documentation. The function does have a doc string.

If you are satisfied with this, you could merge this or I could write the triangular aperture class and then we could merge this. But I figure latter and this combined would be too much code for one pull request.

hamogu:
Each pull request should be a functional unit; if a class has no direct use, but becomes only useful together with another class, then they should be part of the same pull request. There is no upper or lower limit to the size of a PR. We've had PRs where somebody just quickly corrected a few typos in the docs and PRs that contain several weeks of work and a complicated new class with many features.

I have a more general question: Does a "triangular" aperture make sense on its own at all? Presumably, we also want a quadratic, rectangular, five-edged, and hexagonal aperture.
Would it not make more sense to write a general `PolygonAperture` that allows any number of edges?
With a convenience class `RegularPolygonAperture` where all edges have the same length because that's the most common use case?

@bsipocz : What do you think? (You listed polygons in #131. Personally, I've never used anything other than a circular aperture.)

cheekujodhpur:
I was thinking of defining a Polygon as a composition of triangles. And hence I was building up to a triangular aperture class.

`PolygonAperture` will essentially call the methods for `TriangularAperture`.

I realise we are losing data on sub pixel resolution when we are composing a polygon with triangles but with `use_exact=1`, we would not have to worry about that. That's my argument.

bsipocz:
There are several issues with this approach. 
From the beginning I wasn't sure that cutting up a polygon to triangles was the right way to do it. Rectangular apertures are special cases and are used on their own, but noone requested to have triangular ones. It would be good not to go down the path of having all the special cases of triangles/rectangles/pentagons/hexagons/..../, but only have one and only complex, arbitrary polygon class. So while we need a polygon class I can be OK with the idea of basing it on the triangular one until a better implementation is written.

But: I wouldn't merge anything half baked, a user facing API, and the aperture class needs to be put into this PR. We don't have functions that neither have a public API nor used in a public facing function/class etc, and I would stick with that rule. Keeping the codebase of any project is quite a task on it's own, and no need to build in cruft that may or may not become functional code at some point.

Also, any new feature needs to have narrative and docstring documentation as well as tests to be attached.

The other thing is that there were quite a lot of expansion and improvement on the new ``regions`` affiliated package during the last week. Basically it was decided that the core of the aperture photometry (the geometry and masks) should be moved over there and photutils will only keep the photometry related methods (basically the flux summing) and will use ``regions`` it as a dependency. 

Thus I'm cc-ing @astrofrog and @keflavich. Do you think this should go straight to ``regions``?

bsipocz:
@hamogu - I'm a stellar astronomer, too, thus normally using only circular apertures for light curve generation. But polygons were requested by several people, so there must be valid use cases for them.

cheekujodhpur:
If that is your opinion, I will try to complete the Polygon aperture class
and push by tomorrow. I am not comfortable with keeping this and waiting
until a better implementation.

bsipocz:
@cheekujodhpur - I'm glad to hear that you are willing to finish this up. However don't feel being rushed, this don't need to be done by tomorrow. E.g. I'm travelling and won't be able to do any code review and much commenting this week anyway.
Also it may be better to wait for the comments from the people behind ``regions`` before putting in more work into this.

cheekujodhpur:
That sounds fine. 

But: I already had code for polygon aperture. For a high number of sub samples within a pixel, this is tooo slow which is primarily the reason I started with going for triangles. Though I haven't seen people use a very high sub-pixel resolution. So, I have pushed the code for Polygon Overlap Grid for now with tests.

I will hold now before implementing the Aperture class further.

Thanks

astrofrog:
@bsipocz - yes, I think this should go straight to the regions package - I'll provide some links later to show which files this should go in

cheekujodhpur:
@astrofrog What would that basically mean for me? What would I have to do?

hamogu:
@cheekujodhpur : You should take your code and open a new PR in the regions package (astrofrog will tell you where to start). Then we'll close this PR an continue the discussion there and hopefully more it into that package soon.

larrybradley:
@cheekujodhpur I'm going to close this PR.  If you are still interested, please submit a PR to the regions package (https://github.com/astropy/regions).  The plan is that `photutils` will use the `regions` package (when ready) for defining aperture shapes.

PR #336, opened 2016-03-16
Use discretize_model to perform PSF -> PRF translation

### LINKED ISSUE #334: Use astropy.convolution.discretize_model to implement PRFAdapter
The `PRFAdapter` class currently does a quick-and-dirty integration to integrate over pixels.  The `astropy.convolution.discretize_model` is a much more featureful version of this.  So `PRFAdapter` should be changed to use that instead.

stbnps:
Fixes #334

Currently PRFAdapter uses dblquad directly to perform pixel wise integration.

Currently, the PSF -> PRF translation is performed by integrating every pixel, but the user may want a cheaper approach.

PRFAdapter will now use discretize_model to let the user choose how this PSF -> PRF translation takes place.

### DIFF
--- a/photutils/psf.py
+++ b/photutils/psf.py
@@ -18,6 +18,8 @@ from astropy.nddata.utils import add_array, subpixel_indices
 from .extern.nddata_compat import extract_array
 from astropy.nddata import support_nddata
 from astropy.utils.exceptions import AstropyUserWarning
+from astropy.convolution import discretize_model
+
 
 __all__ = ['DiscretePRF', 'IntegratedGaussianPRF', 'PRFAdapter',
            'psf_photometry', 'subtract_psf', 'prepare_psf_model']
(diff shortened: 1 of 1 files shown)


stbnps:
If the evaluation points are monotonic (eg: created using meshgrid) the section:
```
minx = np.min(dx)
miny = np.min(dy)
maxx = np.max(dx) + 0.5
maxy = np.max(dy) + 0.5
```

Could be changed to something like:

```
xbounds = [dx[0, 0], dx[dx.shape[0] - 1, dx.shape[1] - 1]
ybounds = [dy[0, 0], dy[dy.shape[0] - 1, dy.shape[1] - 1]
minx = np.min(xbounds)
miny = np.min(ybounds)
maxx = np.max(xbounds) + 0.5
maxy = np.max(ybounds) + 0.5
```
This way it doesn't search the whole array.

eteq:
@stbnps - this is a good start, thanks!  But I meant a bit more than just replacing the integration machinery with the similar machinery in `discretize_model`.  `discretize_model` has modes other than "integrate", which are much faster.  So there should be a way for the user to tell `discretize_model` what mode to use, and any other parameters that might be necessary (currently the only one is ``factor``, but future versions of astropy might include more ways).  Does that all make sense?

Two other thing: First, you should always provide a description in a new pull request.  Right now you have only the title, and then a comment on a specific detail.  When issuing a new pull request you should try to at least give a little context so that the reviewers know why you're doing what you're doing.  Second, you should reference the underlying issue in the PR itself.  So for example, in the description you might say ``This pull request is meant to fix #334``.

PR #343, opened 2016-03-24
Draft  body for nonparametric morphology module #PyAstro16

astroferreira:
While talking to @eteq in the beginning of Python in Astronomy 2016 I came up with the idea of
creating a module with functionalities for measuring non-parametric morphology. I don't know exactly
where it would fit, but it will need to use a aperture photometry to find Petrosian regions in the future, so in principle it would be better to go in photutils. 

This is a working in progress and I provided the Gini Coefficient as an example for the general structure of the following functionalities that we could implement in the future. I would love to have some feedback in the direction to go from here. 

I found that `photutils.morphology` already exists and all the functionalities in there does not seem to match those for nonparametric morphology or galaxy morphology in general, so I decided to go for a new module, `photutils.morphometry`, that would be easy to grow by itself and have its own signature.

Any inputs?


larrybradley:
Thanks, @astroferreira.  I think this is a good addition to `photutils`.

`photutils.morphometry` is fine for now for these non-parametric morphologies.  Most of the stuff in `photutils.morphology` is about centroids, so we could possible move some of that if you want these tools to go in there.  It doesn't really matter too much though, since most of these tools are available from the main `photutils` namespace.

astroferreira:
Ok @larrybradley, I will keep this work under `photutils.morphometry` then, I was just afraid to generate some confusion about having two 'morpho' modules within `photutils`, but I guess that could be avoided with proper documentation. Thanks!

larrybradley:
Thanks, @astroferreira.  This PR was merged in 2456e5aace57dbdcf07d5ce1e14ef5f19fbcc795 (with some minor changes).  There is now a `photutils.morphology` subpackage and a `photutils.morphology.non_parametric` module if you want to add more functionality.

PR #370, opened 2016-06-10
Add general Background classes

larrybradley:
This large PR contains the following changes:

1.  Turns the `background.py` module into a `background` subpackage.

2.  Adds many new classes for computing a (scalar) background value or background rms value from a (possibly masked) array of any dimension:

  * `MeanBackground`
  * `MedianBackground`
  * `MMMBackground`
  * `SExtractorBackground`
  * `BiweightLocationBackground`
  * `StdBackgroundRMS`
  * `MADStdBackgroundRMS`
  * `BiweightMidvarianceBackgroundRMS`

  Sigma-clipping is performed by default, but can be turned off by setting `sigclip = False`.

3. The existing `BackgroundBase`, `Background` and `BackgroundIDW` classes are now called `BackgroundBase2D`, `Background2D`, and `BackgroundIDW2D`, respectively, since they return 2D background *images*.

My plan is to submit another PR to integrate the general background classes into the 2D background classes (with a similar interface).

Example
------------
```
from photutils import MMMBackground
bkg = MMMBackground(sigma=3., iters=10)
bkg_value = bkg.calc_background(data)
``` 

The classes are designed to be used as an object-oriented interface for other functions, e.g.:
```
bkg = MMMBackground(sigma_lower=5.5, sigma_upper=3., iters=10)
result = psf_photometry(data, ... , background_sub=bkg)
```

in function form (requires `calc_background` -> `__call__`):
```
result = psf_photometry(data, ... , background_sub=np.ma.mean)
```
where the user can define custom background classes if desired.

cc: @eteq, @hcferguson


larrybradley:
The background classes are now callable.  The `__call__` method simply passes the `data` to the `calc_background()` method.

docs/photutils/background.rst
```diff
@@ -158,13 +158,13 @@ mesh of a grid that covers the input data to create a low-resolution
 background map.  The final background or background rms map can then
 be generated by interpolating the low-resolution map.
```
eteq: I guess this comment really belongs just a bit further up... But I wonder if you should add a short example in the previous section that demos all of the various Background (not 2D) classes?  Right now it all uses the sigma clipping functions, but now you can insert all of the various estimators implemented here.  Maybe this could just be a (logarithmic) histogram of the pixel values and then mark the output of the estimators as vertical lines?

photutils/background/core.py
```diff
@@ -0,0 +1,675 @@
...
+            return sigma_clip(data, sig=self.sigma,
+                              iters=self.iters)
+        else:
+            return sigma_clip(data, sigma=self.sigma,
+                              sigma_lower=self.sigma_lower,
+                              sigma_upper=self.sigma_upper,
+                              iters=self.iters)
+
+
+class BackgroundBase(BkgBase):
```
eteq: I think `BackgroundBase` and `BkgBase` are a confusing pair.  I get the intended purpose, but I think it's a bit confusing as it stands now.  I'll leave a general comment about what I think might be a better herirarchy, but even if you don't like that, this name should change.  Maybe `BackgroundBaseValue` and `BackgroundBase`?  Or `BackgroundBaseCentral`?

photutils/background/core.py
```diff
@@ -0,0 +1,675 @@
...
+
+    Alternatively, the background value can be calculated by calling the
+    class instance as a function, e.g.:
+
+    >>> bkg_value = bkg(data)
+    >>> print(bkg_value)    # doctest: +FLOAT_CMP
+    49.5
+    """
+
+    def __init__(self, **kwargs):
```
eteq: This isn't really necessary, right?  That is, if you don't override the `__init__` at all it will just do this?

eteq: (ditto with many of the classes below)

photutils/background/core.py
```diff
@@ -0,0 +1,675 @@
...
+
+        super(MeanBackground, self).__init__(**kwargs)
+
+    def calc_background(self, data):
+
+        if self.sigclip:
+            data = self.sigma_clip(data)
```
eteq: (ditto with many of the classes below)

photutils/background/core.py
```diff
@@ -0,0 +1,675 @@
...
+        super(MedianBackground, self).__init__(**kwargs)
+
+    def calc_background(self, data):
+
+        if self.sigclip:
+            data = self.sigma_clip(data)
+        return np.ma.median(data)
+
+
+class MMMBackground(BackgroundBase):
```
eteq: This is almost identical to `SExtractorBackground` - and I know of other people who prefer slightly different factors to go with median and mean.  Maybe adapt this into somthing like `ModeEstimatorBackground` but with the 3/2 as *parameters*?  Then `MMMBackground` and `SExtractorBackground` can be simple classes along the line of:

```python
class MMMBackground(ModeEstimatorBackground):
    def __init__(self, **kwargs):
        kwargs['median_factor'] = 3
        kwargs['mean_factor'] = 2
        super(MMMBackground, self).__init__(**kwargs)

class SExtractorBackground(ModeEstimatorBackground):
    def __init__(self, **kwargs):
        kwargs['median_factor'] = 2.5
        kwargs['mean_factor'] = 1.5
        super(MMMBackground, self).__init__(**kwargs)

eteq:
@larrybradley - left some inline comments above, and I've got one more general suggestion:
I found the base hierarchy at first conceptually confusing - that is, why is there a `BkgBase` *at all*, when it didn't seem to do anything related to background.  After some thought, I realized it's there because it provides sigma-clipping... But I think that doesn't actually need to be the base of the background heirarchy... And in fact it's a bit confusing because it implies background subtraction *must* involve sigma clipping.

So I have an alternate suggestion that I think is just a small perturbation on what you've done already:  change `BkgBase` to something like `SigmaClipper`, and use it as a *mixin*, rather than the base.  That is, have `BackgroundBase` and `BackgroundRMSBase` be `object` subclasses, but have all the other classes have inheritance signatures along the lines of `MMMBackground(BackgroundBase, SigmaClipper)`.  You might have to be a bit more careful with the super calls, but I *think* it's a non-issue here because there's no `BackgroundBase.__init__` anyway.

docs/photutils/background.rst
larrybradley: Yes, I didn't add the new Background classes to narrative docs yet.  I wanted to wait until the API was firmly decided.

photutils/background/core.py
larrybradley: Done.  I turned `BkgBase` into a `SigmaClip` mixin.

photutils/background/core.py
larrybradley: Done.

photutils/background/core.py
larrybradley: Yes.  I removed the unnecessary `__init__`s, but I worry a bit about about a new user trying to make their own custom class by blindly starting from a copy of one of these classes and not knowing that they need to define an `__init__` (if they don't use the `SigmaClip` mixin, which is where the `__init__` is coming from in most cases).

photutils/background/core.py
larrybradley: I added a `ModeEstimatorBackground` class, with `MMMBackground` as a subclass.  However, `SExtractorBackground` is more complicated than simply ``2.5 * medan() - 1.5 * mean()``, so I did not use `ModeEstimatorBackground` there.

larrybradley:
Thanks, @eteq!  Please review the updates.

eteq:
looks great to me now... except for the test failures.  Any idea what might be specific about numpy 1.9 that would cause this?

larrybradley:
@eteq Thanks.

I guess you missed the ongoing saga with `conda` and `ci-helpers` [e.g. 1, 2, 3].  The very short summary is that `conda 4.1.0` broke a lot of things last week (among other things leaving out the `default` channel).  Our `conda` channels are in a state of flux.  I think the plan was to wipe `astropy-ci-extras` and start over in `conda-channel-ci-extras`, but nothing is in there yet.  `travis-ci` tests are failing across all `astropy` packages, including the core package. `conda` 4.1.1 and 4.1.2 are out now, but I think we need to re-build `astropy` with older `numpy` versions.

[1] https://github.com/astropy/ci-helpers/pull/97
[2] https://github.com/astropy/ci-helpers/issues/98
[3] https://github.com/astropy/ci-helpers/issues/100

eteq:
Ohhh, gotcha - I was confused by the fact that all of the *other* versions still work, but I see now from those ci issues that 1.9 is indeed problematic.

But given that the *others* passed, I would think this is good to merge (assuming you're done making changes, @larrybradley ?)

PR #385, opened 2016-07-02
PSF Photometry in Crowded Fields

mirca:
Hi all,

This is a proposal for a general purpose PSF Photometry module.

The goal is that users may be able to use this module to create their own custom psf photometry, i.e., with any finding, grouping, background subtraction implementations.

For now, we are implementing a version of the NSTAR algorithm which is used in DAOPHOT.

Firstly, we would like to hear feedback from the community on the interface. See this simple gist to get an idea of the interface: 

```
# create a fittable 2d model
my_psf_model = IntegratedGaussianPRF(sigma=3.0)

# create a finder object
my_finder = DAOStarFinder(threshold=5., fwhm=3.0*gaussian_sigma_to_fwhm)

# create a grouping object
my_grouping = DAOGroupStars(crit_separation=1.5*3.0*gaussian_sigma_to_fwhm)

# create a background object
my_bkg = MMMBackground(sigma=3., iters=10)

# create a fitter object
my_fitter = LevMarLSQFitter()

# create psf photometry object
my_photometry = NStarPSFPhotometry(find=my_finder, group=my_group,
                                   bkg=my_bkg, psf_model=my_psf_model,
                                   fitter=my_fitter, niters=5, fitshape=(7,7))

# get photometry results
photometry_results, residual_image = my_photometry(image) 
```

In addition, one of our concerns is whether all these validations of the attributes of the `NStarPSFPhotometry` class are really necessary. Clearly, this prevents many troubles on the user and the developer sides, but also prevents duck-typing and turn things a bit more complicated. (overkill?)

Suggestions are more than welcome.

cc @hamogu @bsipocz @eteq @larrybradley

### DIFF
--- a/photutils/detection/findstars.py
+++ b/photutils/detection/findstars.py
@@ -24,8 +24,8 @@ from .core import find_peaks
 from ..utils.convolution import filter_data
 
 
-__all__ = ['DAOStarFinder', 'IRAFStarFinder', 'StarFinderBase',
-           'daofind', 'irafstarfind']
+__all__ = ['DAOStarFinder', 'IRAFStarFinder', 'StarFinderBase', 'daofind',
+           'irafstarfind']
 
 
 class _ABCMetaAndInheritDocstrings(InheritDocstrings, abc.ABCMeta):
@@ -326,6 +326,61 @@ class IRAFStarFinder(StarFinderBase):
         self.exclude_border = exclude_border
 
     def find_stars(self, data):
+        """
+        Parameters
+        ----------
+        data : array_like
+            The 2D array of the image.
+        
+        Returns
+        -------
+        table : `~astropy.table.Table`
+
+            A table of found objects with the following parameters:
+
+            * ``id``: unique object identification number.
+            * ``xcentroid, ycentroid``: object centroid (zero-based origin).
+            * ``fwhm``: estimate of object FWHM from image moments.
+            * ``sharpness``: object sharpness calculated from image moments.
+            * ``roundness``: object ellipticity calculated from image moments.
+            * ``pa``:  object position angle in degrees from the positive x
+              axis calculated from image moments.
+            * ``npix``: number of pixels in the object used to calculate
+              ``flux``.
+            * ``sky``: the derived background sky value, unless ``sky`` was
+              input.  If ``sky`` was input, then that value overrides the
+              background sky estimation.
+            * ``peak``: the peak, sky-subtracted, pixel value of the object.
+            * ``flux``: the object sky-subtracted flux, calculated by
+              summing object pixels over the Gaussian kernel.  The
+              derivation matches that of `starfind`_ if ``sky`` is ``None``.
+            * ``mag``: the object instrumental magnitude calculated as
+              ``-2.5 * log10(flux)``.  The derivation matches that of
+              `starfind`_ if ``sky`` is ``None``.
+        
+        Notes
+        -----
+        For the convolution step, this routine sets pixels beyond the image
+        borders to 0.0.  The equivalent parameters in `starfind`_ are
+        ``boundary='constant'`` and ``constant=0.0``.
+
+        IRAF's `starfind`_ uses ``hwhmpsf``, ``fradius``, and ``sepmin`` as
+        input parameters.  The equivalent input values for ``IRAFStarFinder``
+        are:
+
+        * ``fwhm = hwhmpsf * 2``
+        * ``sigma_radius = fradius * sqrt(2.0*log(2.0))``
+        * ``minsep_fwhm = 0.5 * sepmin``
+
+        The main differences between ``daofind`` and ``irafstarfind`` are:
+
+        * ``IRAFStarFinder`` always uses a 2D circular Gaussian kernel,
+          while ``DAOStarFinder`` can use an elliptical Gaussian kernel.
+
+        * ``IRAFStarFinder`` calculates the objects' centroid, roundness,
+          and sharpness using image moments.
+        """
+
         starfind_kernel = _FindObjKernel(self.fwhm, ratio=1.0, theta=0.0,
                                          sigma_radius=self.sigma_radius)
         min_separation = max(2, int((self.fwhm * self.minsep_fwhm) + 0.5))
--- a/photutils/detection/findstars.py
+++ b/photutils/detection/findstars.py
@@ -24,8 +24,8 @@ from .core import find_peaks
 from ..utils.convolution import filter_data
 
 
-__all__ = ['DAOStarFinder', 'IRAFStarFinder', 'StarFinderBase', 'daofind',
-           'irafstarfind']
+__all__ = ['DAOStarFinder', 'IRAFStarFinder', 'StarFinderBase',
+           'daofind', 'irafstarfind']
 
 
 class _ABCMetaAndInheritDocstrings(InheritDocstrings, abc.ABCMeta):
@@ -326,61 +326,6 @@ class IRAFStarFinder(StarFinderBase):
         self.exclude_border = exclude_border
 
     def find_stars(self, data):
-        """
-        Parameters
-        ----------
-        data : array_like
-            The 2D array of the image.
-        
-        Returns
-        -------
-        table : `~astropy.table.Table`
-
-            A table of found objects with the following parameters:
-
-            * ``id``: unique object identification number.
-            * ``xcentroid, ycentroid``: object centroid (zero-based origin).
-            * ``fwhm``: estimate of object FWHM from image moments.
-            * ``sharpness``: object sharpness calculated from image moments.
-            * ``roundness``: object ellipticity calculated from image moments.
-            * ``pa``:  object position angle in degrees from the positive x
-              axis calculated from image moments.
-            * ``npix``: number of pixels in the object used to calculate
-              ``flux``.
-            * ``sky``: the derived background sky value, unless ``sky`` was
-              input.  If ``sky`` was input, then that value overrides the
-              background sky estimation.
-            * ``peak``: the peak, sky-subtracted, pixel value of the object.
-            * ``flux``: the object sky-subtracted flux, calculated by
-              summing object pixels over the Gaussian kernel.  The
-              derivation matches that of `starfind`_ if ``sky`` is ``None``.
-            * ``mag``: the object instrumental magnitude calculated as
-              ``-2.5 * log10(flux)``.  The derivation matches that of
-              `starfind`_ if ``sky`` is ``None``.
-        
-        Notes
-        -----
-        For the convolution step, this routine sets pixels beyond the image
-        borders to 0.0.  The equivalent parameters in `starfind`_ are
-        ``boundary='constant'`` and ``constant=0.0``.
-
-        IRAF's `starfind`_ uses ``hwhmpsf``, ``fradius``, and ``sepmin`` as
-        input parameters.  The equivalent input values for ``IRAFStarFinder``
-        are:
-
-        * ``fwhm = hwhmpsf * 2``
-        * ``sigma_radius = fradius * sqrt(2.0*log(2.0))``
-        * ``minsep_fwhm = 0.5 * sepmin``
-
-        The main differences between ``daofind`` and ``irafstarfind`` are:
-
-        * ``IRAFStarFinder`` always uses a 2D circular Gaussian kernel,
-          while ``DAOStarFinder`` can use an elliptical Gaussian kernel.
-
-        * ``IRAFStarFinder`` calculates the objects' centroid, roundness,
-          and sharpness using image moments.
-        """
-
         starfind_kernel = _FindObjKernel(self.fwhm, ratio=1.0, theta=0.0,
                                          sigma_radius=self.sigma_radius)
         min_separation = max(2, int((self.fwhm * self.minsep_fwhm) + 0.5))
--- a/photutils/psf/__init__.py
+++ b/photutils/psf/__init__.py
@@ -10,3 +10,4 @@ from .matching import *
 from .models import *
 from .psfphotometry import *
 from .groupstars import *
+from .photometry import *
--- a/photutils/psf/__init__.py
+++ b/photutils/psf/__init__.py
@@ -8,6 +8,5 @@ from .funcs import *
 from .groupstars import *
 from .matching import *
 from .models import *
-from .psfphotometry import *
 from .groupstars import *
 from .photometry import *
--- a/photutils/psf/__init__.py
+++ b/photutils/psf/__init__.py
@@ -8,3 +8,5 @@ from .funcs import *
 from .groupstars import *
 from .matching import *
 from .models import *
+from .psfphotometry import *
+from .groupstars import *
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -2,9 +2,7 @@
 """Module which provides classes to perform PSF Photometry"""
 
 from __future__ import division
-import abc
 import numpy as np
-from astropy.extern import six
 from astropy.table import Table, vstack
 from astropy.modeling.fitting import LevMarLSQFitter
 from astropy.nddata.utils import overlap_slices, NoOverlapError
@@ -184,12 +182,9 @@ class DAOPhotPSFPhotometry(object):
             and the original image.
         """
         
-        if positions is None:
-            return self.do_photometry(image)
-        else:
-            return self.do_fixed_photometry(image, positions)
+        return self.do_photometry(image, positions)
 
-    def do_photometry(self, image):
+    def do_photometry(self, image, positions=None):
         """
         Perform PSF photometry in ``image``. This method assumes that
         ``psf`` has centroids and flux parameters which will be fitted to the
@@ -197,12 +192,21 @@ class DAOPhotPSFPhotometry(object):
         ``psf``, will be fitted to groups of starts automatically identified
         by ``group``. Also, ``image`` is not assumed to be background
         subtracted.
+        If positions are not ``None`` then this method performs forced PSF
+        photometry. If the centroid positions are set
+        as ``fixed`` in the PSF model ``psf``, then the optimizer will only
+        consider the flux as a variable. Otherwise, ``positions`` will be used
+        as initial guesses for the centroids.
 
         Parameters
         ----------
         image : 2D array-like, `~astropy.io.fits.ImageHDU`,
         `~astropy.io.fits.HDUList`
             Image to perform photometry.
+        positions: `~astropy.table.Table` (default=None)
+            Positions (in pixel coordinates) at which to *start* the fit for
+            each object. Columns 'x_0' and 'y_0' must be present.
+            'flux_0' can also be provided to set initial fluxes.
         
         Returns
         -------
@@ -214,98 +218,66 @@ class DAOPhotPSFPhotometry(object):
             Residual image calculated by subtracting the fitted sources
             and the original image.
         """
-
-        outtab = Table([[], [], [], [], [], []],
-                       names=('id', 'group_id', 'x_fit', 'y_fit', 'flux_fit',
-                              'iter_detected'),
-                       dtype=('i4', 'i4', 'f8', 'f8', 'f8', 'i4'))
-
+        
         residual_image = image.copy()
         residual_image = residual_image - self.bkg(image)
-        sources = self.find(residual_image)
         
         if self.aperture_radius is None:
             if hasattr(self.psf, 'fwhm'):
                 self.aperture_radius = self.psf.fwhm.value
             elif hasattr(self.psf, 'sigma'):
-                self.aperture_radius = self.psf.sigma.value*gaussian_sigma_to_fwhm
+                self.aperture_radius = self.psf.sigma.value*\
+                                       gaussian_sigma_to_fwhm
 
-        apertures = CircularAperture((sources['xcentroid'],
-                                      sources['ycentroid']),
-                                     r=self.aperture_radius)
-
-        sources['aperture_flux'] = aperture_photometry(residual_image,
-                                              apertures)['aperture_sum']
-        n = 1
-        while(n <= self.niters and len(sources) > 0):
-            intab = Table(names=['id', 'x_0', 'y_0', 'flux_0'],
-                          data=[sources['id'], sources['xcentroid'],
-                          sources['ycentroid'], sources['aperture_flux']])
-            star_groups = self.group(intab)
-            tab, residual_image = self.nstar(residual_image, star_groups)
-            tab['iter_detected'] = n*np.ones(tab['x_fit'].shape, dtype=np.int)
-            outtab = vstack([outtab, tab])
+        if positions is None:
+            outtab = Table([[], [], [], [], [], []],
+                           names=('id', 'group_id', 'x_fit', 'y_fit',
+                                  'flux_fit', 'iter_detected'),
+                           dtype=('i4', 'i4', 'f8', 'f8', 'f8', 'i4'))
+            
             sources = self.find(residual_image)
-
-            if len(sources) > 0:
-                apertures = CircularAperture((sources['xcentroid'],
-                                              sources['ycentroid']),
-                                             r=self.aperture_radius)
-                sources['flux'] = aperture_photometry(residual_image,
-                                                      apertures)['aperture_sum']
-            n += 1
-        return outtab, residual_image
-
-    def do_fixed_photometry(self, image, positions):
-        """
-        Perform PSF photometry for the case that the centroid positions of the
-        starts are known with high accuracy. If the centroid positions are set
-        as ``fixed`` in the PSF model ``psf``, then the optimizer will only
-        consider the flux as a variable. Otherwise, ``positions`` will be used
-        as initial guesses for the centroids. Also, ``image`` is not assumed
-        to be background subtracted.
-
-        Parameters
-        ----------
-        image : 2D array-like, `~astropy.io.fits.ImageHDU`,
-        `~astropy.io.fits.HDUList`
-            Image to perform photometry
-        positions: `~astropy.table.Table`
-            Positions (in pixel coordinates) at which to *start* the fit for
-            each object. Columns 'x_0' and 'y_0' must be present.
-            'flux_0' can also be provided to set initial fluxes.
-
-        Returns
-        -------
-        outtab : `~astropy.table.Table`
-            Table with the photometry results, i.e., centroids and fluxes
-            estimations.
-        residual_image : array-like, `~astropy.io.fits.ImageHDU`,
-        `~astropy.io.fits.HDUList`
-            Residual image calculated by subtracting the fitted sources
-            from the original image.
-        """
-
-        residual_image = image.copy()
-        residual_image = residual_image - self.bkg(image)
-
-        if 'flux_0' not in positions.colnames:
-            if self.aperture_radius is None:
-                 self.aperture_radius = self.psf.sigma.value*\
-                                        gaussian_sigma_to_fwhm
-            apertures = CircularAperture((positions['x_0'], positions['y_0']),
+            
+            apertures = CircularAperture((sources['xcentroid'],
+                                          sources['ycentroid']),
                                          r=self.aperture_radius)
 
-            positions['aperture_flux'] = aperture_photometry(residual_image,\
-                                         apertures)['aperture_sum']
+            sources['aperture_flux'] = aperture_photometry(residual_image,
+                                                  apertures)['aperture_sum']
+            n = 1
+            while(n <= self.niters and len(sources) > 0):
+                intab = Table(names=['id', 'x_0', 'y_0', 'flux_0'],
+                              data=[sources['id'], sources['xcentroid'],
+                              sources['ycentroid'], sources['aperture_flux']])
+                star_groups = self.group(intab)
+                tab, residual_image = self.nstar(residual_image, star_groups)
+                tab['iter_detected'] = n*np.ones(tab['x_fit'].shape,
+                                                 dtype=np.int)
+                outtab = vstack([outtab, tab])
+                sources = self.find(residual_image)
 
-        intab = Table(names=['x_0', 'y_0', 'flux_0'],
-                      data=[positions['x_0'], positions['y_0'],
-                      positions['aperture_flux']])
+                if len(sources) > 0:
+                    apertures = CircularAperture((sources['xcentroid'],
+                                                  sources['ycentroid']),
+                                                 r=self.aperture_radius)
+                    sources['flux'] = aperture_photometry(residual_image,\
+                                                apertures)['aperture_sum']
+                n += 1
+        else:
+            if 'flux_0' not in positions.colnames:
+                apertures = CircularAperture((positions['x_0'],
+                                              positions['y_0']),
+                                             r=self.aperture_radius)
 
-        star_groups = self.group(intab)
-        outtab, residual_image = self.nstar(residual_image, star_groups)
-        
+                positions['flux_0'] = aperture_photometry(residual_image,\
+                                             apertures)['aperture_sum']
+
+            intab = Table(names=['x_0', 'y_0', 'flux_0'],
+                          data=[positions['x_0'], positions['y_0'],
+                          positions['flux_0']])
+
+            star_groups = self.group(intab)
+            outtab, residual_image = self.nstar(residual_image, star_groups)
+
         return outtab, residual_image
 
     def get_uncertainties(self):
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -8,19 +8,15 @@ from astropy.extern import six
 from astropy.table import Table, vstack
 from astropy.modeling.fitting import LevMarLSQFitter
 from astropy.nddata.utils import overlap_slices, NoOverlapError
+from astropy.stats import gaussian_sigma_to_fwhm
 from ..psf import subtract_psf
+from ..aperture import CircularAperture, aperture_photometry
 
 
-__all__ = ['PSFPhotometryBase', 'DAOPhotPSFPhotometry']
+__all__ = ['DAOPhotPSFPhotometry']
 
 
-@six.add_metaclass(abc.ABCMeta)
-class PSFPhotometryBase(object):
-    @abc.abstractmethod
-    def do_photometry(self, image):
-        pass
-
-class DAOPhotPSFPhotometry(PSFPhotometryBase):
+class DAOPhotPSFPhotometry(object):
     """
     This class implements the DAOPHOT algorithm proposed by Stetson
     (1987) to perform point spread function photometry in crowded fields,
@@ -30,7 +26,7 @@ class DAOPhotPSFPhotometry(PSFPhotometryBase):
     """
 
     def __init__(self, group, bkg, psf, fitshape, find=None,
-                 fitter=LevMarLSQFitter(), niters=3):
+                 fitter=LevMarLSQFitter(), niters=3, aperture_radius=None):
         """
         Attributes
         ----------
@@ -67,16 +63,10 @@ class DAOPhotPSFPhotometry(PSFPhotometryBase):
             This object needs to identify three parameters (position of
             center in x and y coordinates and the flux) in order to set them
             to suitable starting values for each fit. The names of these
-            parameters can be given as follows:
-
-            - Set ``psf.psf_xname``, ``psf.psf_yname`` and
-              ``psf.psf_fluxname`` to strings with the names of the respective
-              psf model parameter.
-            - If those attributes are not found, the names ``x_0``, ``y_0``
-              and ``flux`` are assumed.
+            parameters should be given as ``x_0``, ``y_0`` and ``flux``.
 
             `~photutils.psf.prepare_psf_model` can be used to prepare any 2D
-            model to match these assumptions.
+            model to match this assumption.
         fitshape : array-like
             Rectangular shape around the center of a star which will be used
             to collect the data to do the fitting, e.g. (5, 5) means to take
@@ -89,6 +79,9 @@ class DAOPhotPSFPhotometry(PSFPhotometryBase):
         niters : int (default=3)
             Number of iterations to perform the loop FIND, GROUP, SUBTRACT,
             NSTAR.
+        aperture_radius : float (default=None)
+            The radius (in units of pixels) used to compute initial estimates
+            for the fluxes of sources. If ``None``, one fwhm will be used. 
 
         Notes
         -----
@@ -112,50 +105,69 @@ class DAOPhotPSFPhotometry(PSFPhotometryBase):
         self.fitter = fitter
         self.niters = niters
         self.fitshape = fitshape
+        self.aperture_radius = aperture_radius
 
-        @property
-        def niters(self):
-            return self._niters
+    @property
+    def niters(self):
+        return self._niters
 
-        @niters.setter
-        def niters(self, niters):
-            if isinstance(niters, int) and niters > 0:
-                self._niters = niters
+    @niters.setter
+    def niters(self, value):
+        try:
+            if value <= 0:
+                raise ValueError('niters must be positive.')
             else:
-                raise ValueError('niters must be an interger-valued number, '
-                                 'received niters = {}'.format(niters))
-        
-        @property
-        def fitshape(self):
-            return self._fitshape
+                self._niters = int(value)
+        except:
+            raise ValueError('niters must be an integer or convertable '
+                             'into an integer.')
+    
+    @property
+    def fitshape(self):
+        return self._fitshape
 
-        @fitshape.setter
-        def fitshape(self, fitshape):
-            fitshape = np.asarray(fitshape)
-            if len(fitshape) == 2:
-                if np.all(fitshape) > 0:
-                    if np.all(fitshape) % 2 == 1:
-                        self._fitshape = fitshape
-                    else:
-                        raise ValueError('fitshape must be odd '
-                                         'integer-valued, '
-                                         'received fitshape = {}'\
-                                         .format(fitshape))
+    @fitshape.setter
+    def fitshape(self, value):
+        value = np.asarray(value)
+        if value.size == 2:
+            if np.all(value) > 0:
+                if np.all(value % 2) == 1:
+                    self._fitshape = tuple(value)
                 else:
-                    raise ValueError('fitshape must have positive elements, '
+                    raise ValueError('fitshape must be odd integer-valued, '
                                      'received fitshape = {}'\
-                                     .format(fitshape))
+                                     .format(value))
             else:
-                raise ValueError('fitshape must have two dimensions, '
-                                 'received fitshape = {}'.format(fitshape))
+                raise ValueError('fitshape must have positive elements, '
+                                 'received fitshape = {}'\
+                                 .format(value))
+        else:
+            raise ValueError('fitshape must have two dimensions, '
+                             'received fitshape = {}'.format(value))
 
-    def __call__(self, **kwargs):
+    @property
+    def aperture_radius(self):
+        return self._aperture_radius
+
+    @aperture_radius.setter
+    def aperture_radius(self, value):
+        if isinstance(value, (int, float)) and value > 0:
+            self._aperture_radius = value
+        elif value is None:
+            self._aperture_radius = value
+        else:
+            raise ValueError('aperture_radius must be a real-valued '
+                             'number, received aperture_radius = {}'
+                             .format(value))
+
+
+    def __call__(self, image, positions=None):
         """
         Parameters
         ----------
         image : 2D array-like, `~astropy.io.fits.ImageHDU`,
         `~astropy.io.fits.HDUList`
-            Image to perform photometry
+            Image to perform photometry.
         positions : `~astropy.table.Table` (optional)
             Positions, in pixel coordinates, at which stars are located.
             The columns must be named as 'x_0' and 'y_0'. 'flux_0' can also
@@ -172,18 +184,25 @@ class DAOPhotPSFPhotometry(PSFPhotometryBase):
             and the original image.
         """
         
-        if len(kwargs) == 1:
-            return self.do_photometry(kwargs['image'])
-        elif len(kwargs) == 2:
-            return self.do_fixed_photometry(kwargs['image'], kwargs['positions'])
+        if positions is None:
+            return self.do_photometry(image)
+        else:
+            return self.do_fixed_photometry(image, positions)
 
     def do_photometry(self, image):
         """
+        Perform PSF photometry in ``image``. This method assumes that
+        ``psf`` has centroids and flux parameters which will be fitted to the
+        data provided in ``image``. A compound model, in fact a sum of
+        ``psf``, will be fitted to groups of starts automatically identified
+        by ``group``. Also, ``image`` is not assumed to be background
+        subtracted.
+
         Parameters
         ----------
         image : 2D array-like, `~astropy.io.fits.ImageHDU`,
         `~astropy.io.fits.HDUList`
-            Image to perform photometry
+            Image to perform photometry.
         
         Returns
         -------
@@ -198,36 +217,62 @@ class DAOPhotPSFPhotometry(PSFPhotometryBase):
 
         outtab = Table([[], [], [], [], [], []],
                        names=('id', 'group_id', 'x_fit', 'y_fit', 'flux_fit',
-                            'iter_detected'),
+                              'iter_detected'),
                        dtype=('i4', 'i4', 'f8', 'f8', 'f8', 'i4'))
 
         residual_image = image.copy()
         residual_image = residual_image - self.bkg(image)
         sources = self.find(residual_image)
-       
+        
+        if self.aperture_radius is None:
+            if hasattr(self.psf, 'fwhm'):
+                self.aperture_radius = self.psf.fwhm.value
+            elif hasattr(self.psf, 'sigma'):
+                self.aperture_radius = self.psf.sigma.value*gaussian_sigma_to_fwhm
+
+        apertures = CircularAperture((sources['xcentroid'],
+                                      sources['ycentroid']),
+                                     r=self.aperture_radius)
+
+        sources['aperture_flux'] = aperture_photometry(residual_image,
+                                              apertures)['aperture_sum']
         n = 1
         while(n <= self.niters and len(sources) > 0):
             intab = Table(names=['id', 'x_0', 'y_0', 'flux_0'],
                           data=[sources['id'], sources['xcentroid'],
-                          sources['ycentroid'], sources['flux']])
+                          sources['ycentroid'], sources['aperture_flux']])
             star_groups = self.group(intab)
             tab, residual_image = self.nstar(residual_image, star_groups)
             tab['iter_detected'] = n*np.ones(tab['x_fit'].shape, dtype=np.int)
             outtab = vstack([outtab, tab])
             sources = self.find(residual_image)
+
+            if len(sources) > 0:
+                apertures = CircularAperture((sources['xcentroid'],
+                                              sources['ycentroid']),
+                                             r=self.aperture_radius)
+                sources['flux'] = aperture_photometry(residual_image,
+                                                      apertures)['aperture_sum']
             n += 1
         return outtab, residual_image
 
     def do_fixed_photometry(self, image, positions):
         """
+        Perform PSF photometry for the case that the centroid positions of the
+        starts are known with high accuracy. If the centroid positions are set
+        as ``fixed`` in the PSF model ``psf``, then the optimizer will only
+        consider the flux as a variable. Otherwise, ``positions`` will be used
+        as initial guesses for the centroids. Also, ``image`` is not assumed
+        to be background subtracted.
+
         Parameters
         ----------
         image : 2D array-like, `~astropy.io.fits.ImageHDU`,
         `~astropy.io.fits.HDUList`
             Image to perform photometry
         positions: `~astropy.table.Table`
-            Positions at which to *start* the fit for each object, in pixel
-            coordinates. Columns 'x_0' and 'y_0' must be present.
+            Positions (in pixel coordinates) at which to *start* the fit for
+            each object. Columns 'x_0' and 'y_0' must be present.
             'flux_0' can also be provided to set initial fluxes.
 
         Returns
@@ -244,14 +289,19 @@ class DAOPhotPSFPhotometry(PSFPhotometryBase):
         residual_image = image.copy()
         residual_image = residual_image - self.bkg(image)
 
-        if 'flux_0' in positions.colnames:
-               intab = Table(names=['x_0', 'y_0', 'flux_0'],
-                             data=[positions['x_0'], positions['y_0'],
-                                   positions['flux_0']])
-        else:
-            intab = Table(names=['x_0', 'y_0', 'flux_0'],
-                          data=[positions['x_0'], positions['y_0'],
-                                np.ones(len(positions))])
+        if 'flux_0' not in positions.colnames:
+            if self.aperture_radius is None:
+                 self.aperture_radius = self.psf.sigma.value*\
+                                        gaussian_sigma_to_fwhm
+            apertures = CircularAperture((positions['x_0'], positions['y_0']),
+                                         r=self.aperture_radius)
+
+            positions['aperture_flux'] = aperture_photometry(residual_image,\
+                                         apertures)['aperture_sum']
+
+        intab = Table(names=['x_0', 'y_0', 'flux_0'],
+                      data=[positions['x_0'], positions['y_0'],
+                      positions['aperture_flux']])
 
         star_groups = self.group(intab)
         outtab, residual_image = self.nstar(residual_image, star_groups)
@@ -262,14 +312,14 @@ class DAOPhotPSFPhotometry(PSFPhotometryBase):
         """
         Return the uncertainties on the fitted parameters
         """
-        pass
+        raise NotImplementedError
 
     def nstar(self, image, star_groups):
         """
         Fit, as appropriate, a compound or single model to the given
-        `star_groups`. Groups are fitted sequentially from the smallest to
-        the biggest. In each iteration, `image` is subtracted by the previous
-        fitted group. 
+        ``star_groups``. Groups are fitted sequentially from the smallest to
+        the biggest. In each iteration, ``image`` is subtracted by the
+        previous fitted group.
         
         Parameters
         ----------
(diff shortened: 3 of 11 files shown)


photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,319 @@
...
+        self.niters = niters
+        self.fitshape = fitshape
+
+    @property
+    def find(self):
+        return self._find
+    
+    @find.setter
+    def find(self, find):
+        if isinstance(find, StarFinderBase)
```
bsipocz: Missing ``:`` from the end of line

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,319 @@
...
+    def group(self, group):
+        if isinstance(group, GroupStarsBase):
+            self._group = group
+        else:
+            raise ValueError('group is expected to be an instance of '
+                             'GroupStarsBase, received {}.'.\
+                             format(type(group)))
+    
+    @property
+    def bkg(self)
```
bsipocz: missing ``:``

hamogu: I suggest not to use properties for this here anyway, because it prevents a user for using duck typing.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,319 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+class PSFPhotometryBase(object):
+    __metaclass__ = abc.ABCMeta
```
bsipocz: We tend to use ``@six.add_metaclass``  decorator to cover for both py2 and py3

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""Module which provides classes to perform PSF Photometry"""
+
+from __future__ import division
+import abc, six
```
hamogu: @bsipocz : You always know these things: Should this be an import of "six" (which requires users to have six installed) or "astropy.extern.six" (because users certainly have astropy installed anyway)?

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+class PSFPhotometryBase(object):
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
+    def do_photometry(self):
+        pass
+
+
+class NStarPSFPhotometry(PSFPhotometryBase):
+    """
```
hamogu: I suggest "StetsonPSFPhotometry" or "DaophotPhotometry". I think those names are better known. Also, the docs say "This class basically implements the loop FIND, GROUP, NSTAR, SUBTRACT, FIND", so it does more than just NSTAR and we certainly don't want to call it "FINDGROUPNSTARSUBTRACTFINDLoopPSFPhotometry"  ;-)

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+        find : an instance of any StarFinderBase subclasses
+        group : an instance of any GroupStarsBase subclasses
+        bkg : an instance of any BackgroundBase2D (?) subclasses
+        psf : Fittable2DModel instance
+        fitter : Fitter instance
+        niters : int
+            number of iterations for the loop FIND, GROUP, SUBTRACT, NSTAR
+        fitshape : array-like
+            rectangular shape around the center of a star which will be used
+            to collect the data to do the fitting
```
hamogu: add an example, because it's not clear to me what the format is:
- [width, height]
- [height, width]
- [-delta_x, - delta_y, delta_x, delta_y] = [-5, -2, 5, 2]
- something else?

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+    def __init__(self, find, group, bkg, psf, fitter, niters, fitshape):
+        """
+        Attributes
+        ----------
+        find : an instance of any StarFinderBase subclasses
+        group : an instance of any GroupStarsBase subclasses
+        bkg : an instance of any BackgroundBase2D (?) subclasses
+        psf : Fittable2DModel instance
+        fitter : Fitter instance
+        niters : int
```
hamogu: Maybe niters and fitshape should have default values? I'm not sure what's best. On the one hand, we'd like things to just work, on the other hand, we don't want people to publish results based on defaults that should have been adjusted to their dataset.
Can we at least that `niter` to 3 or so? That will be enough for almost all cases and stops stupidity like n=50000 "because I just want the best results" (OK, it does not stop it, but it signals that `niter` should probably be a relatively small number).

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+
+        @property
+        def niters(self):
+            return self._niters
+        @niters.setter
+        def niters(self, niters):
+            if isinstance(niters, int) and niters > 0:
+                self._niters = niters
+            else:
+                raise ValueError('niters is not defined properly, '
```
hamogu: What's "properly"? Why not be specific and say "must be a positive integer"?

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+        @property
+        def fitshape(self):
+            return self._fitshape
+        @fitshape.setter
+        def fitshape(self, fitshape):
+            fitshape = np.asarray(fitshape)
+            if len(fitshape) == 2 and np.all(fitshape) > 0:
+                self._fitshape = fitshape
+            else:
+                raise ValueError('fitshape is not defined properly, '
```
hamogu: Can fitshape be an even number? If so, what would a lengths of 6 mean? (-2,-1,0,1,2,3)?

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+    def __call__(self, image):
+        """
+        Parameters
+        ----------
+        image : array-like, ImageHDU, HDUList
+            image to perform photometry
+        
+        Returns
+        -------
+        outtab : astropy.table.Table
```
hamogu:
```
`astropy.table.Table`
```
will user intersphinx to generate links to the astropy table documentation.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+            else:
+                raise ValueError('fitshape is not defined properly, '
+                                 'received fitshape = {}'.format(fitshape))
+
+    def __call__(self, image):
+        """
+        Parameters
+        ----------
+        image : array-like, ImageHDU, HDUList
+            image to perform photometry
```
hamogu: Similar to what I said three lines below
```
`~astropy.io.fits.ImageHDU`
```
will generate a link to the ImageHDU docs. (the `~` means " display on the the last part, the IMAGEHDU and not the astropy.io.fits when you render the link in html)

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+        residual_image : array-like, ImageHDU, HDUList
+            Residual image calculated by subtracting the fitted sources
+            and the original image.
+        """
+
+        return self.do_photometry(image)
+
+    def _nstar(self, image, star_groups):
+        """
+        Fit, as appropriate, a compound or single model to a given `groups` of
```
hamogu:
```
`groups`
```
must be 
```
``star_groups``
```
(double backticks to show sphinx that this is not a link)

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+            estimations.
+        residual_image : array-like, ImageHDU, HDUList
+            Residual image calculated by subtracting the fitted sources
+            and the original image.
+        """
+
+        return self.do_photometry(image)
+
+    def _nstar(self, image, star_groups):
+        """
```
hamogu: Why is this a private function? Under usual circumstances we would not want a user to call it directly, but do we really need to to be private?

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+        stars. Groups are fitted sequentially from the smallest to the biggest. In
+        each iteration, `image` is subtracted by the previous fitted group. 
+        
+        Parameters
+        ----------
+        image : numpy.ndarray
+            Background-subtracted image.
+        star_groups : list of `~astropy.table.Table`
+            Each `~astropy.table.Table` in this list corresponds to a group of
+            mutually overlapping starts.
```
hamogu: starts -> stars

hamogu: Each table must contain the following columns: id, ...

hamogu: Actually, I just saw in the code, ti's no longer a list of tables, it's a single table with column group_id.
I strongly prefer this approach (as you now), but the docs have to match that ;-)

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+        result_tab = Table([[], [], [], [], []],
+                           names=('id', 'group_id', 'x_fit', 'y_fit',
+                                  'flux_fit'),
+                           dtype=('i4', 'i4', 'f8', 'f8', 'f8'))
+
+        star_groups = star_groups.group_by('group_id')
+        
+        groups_order = []
+        for g in star_groups.groups:
+            groups_order.append(len(g))
```
hamogu: Does this really matter?
Stetson did smallest to biggest, but I thought the groups are independent, so the order should (?) not matter. If you just use rthe order of "group_id' you a) save a few lines and b) allow the user a way to order groups in any way they like if necessary (by relabelling the colum group_id).
I assume that Stetson did smallest-to-largest because larger groups are more likely to crash the fit. That's still true, but since we assemble the results table in memory (while Stetson writes every line to disk when it's fit) we implicitly require that the code runs all the way through anyway.
Instead, I suggest something like the following sentence in the docstring:

```
Notes
-------
If there are problems with fitting large groups, change the parameters of the grouping algorithm to reduce the number of sources in each group or input a ``star_groups`` table that only includes the groups that are relevant (e.g. manually remove all entries that coincide with artifacts).
```

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+
+        N = len(groups_order)
+        while N > 0:
```
hamogu: `for n in range (N)` ?

photutils/psf/psfphotometry.py
mirca: You're right, it doesn't make much sense to let `fitshape` be even. Thanks!

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+            curr_order = np.min(groups_order)
+            n = 0
+            N = len(groups_order)
```
hamogu: rundundant with line 129.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+            while(n < N):
+                if curr_order == len(star_groups.groups[n]):
+                    group_psf = self._get_sum_psf_model(star_groups.groups[n])
```
hamogu: Ah, I see. I think you can replace the double while ad the if with
```
groups = star_groups.group_by('group_id')
for group in groups.groups:
    group_psf = self._get_sum_psf_model(group)
    ....
```

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+                    x, y, data = self._get_shape_and_data(shape=self.fitshape,\
+                                             star_group=star_groups.groups[n],
+                                                          image=image)
+                    fit_model = self.fitter(group_psf, x, y, data)
+                    print(fit_model)
```
hamogu: I'm to a big fan of unconditional print statements. There is no `verobse=False` flag or similar to switch it off. On the other hand, printing out stuff shows the user that the fit is still going on when the process runs a long time.
What do other people think?

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+        param_tab : ~astropy.table.Table
+            Table that contains the fitted parameters.
+        """
+
+        param_tab = Table([[], [], [], [], []],
+                          names=('id', 'group_id', 'x_fit', 'y_fit',
+                                 'flux_fit'),
+                          dtype=('i4', 'i4', 'f8', 'f8', 'f8'))
+        if np.size(fit_model) == 1:
+            param_tab.add_row([[star_group['id'][0]],
```
hamogu: It's OK to have `x_0_0` etc for a group that has only one star and then you don't need a special case for it.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+            for i in range(np.size(fit_model)):
+                param_tab.add_row([[star_group['id'][i]],
+                                   [star_group['group_id'][i]],
+                                   [getattr(fit_model,'x_0_'+str(i)).value],
+                                   [getattr(fit_model, 'y_0_'+str(i)).value],
+                                   [getattr(fit_model, 'flux_'+str(i)).value]])
+        return param_tab
+
+    def _get_shape_and_data(self, shape, star_group, image):
+        """
```
hamogu: This is very similar to 
https://astropy.readthedocs.io/en/latest/api/astropy.nddata.utils.extract_array.html#astropy.nddata.utils.extract_array
(see also the list of related functions in https://astropy.readthedocs.io/en/latest/nddata/index.html#module-astropy.nddata.utils).
In fact I wrote `extract_array` specifically to do PSF photometry. It was the longest PR I ever had with about 100 comments (see https://github.com/astropy/astropy/pull/3333 - and you thought we are thorough on your pull requests ;-)  ).
Please don't make that wasted by implementig it again here. Also (based on the experiences in astropy #3333) I can tell you that the implementation here will have trouble on the edges, and will not allow me to extract sources whose center is not on the chip (but if they are bright, we can still fit a PSF to the wings and get a reasonable flux, and I have a dataset where I want to do just that).

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+        
+        Returns
+        -------
+        sum_psf : CompoundModel
+            `CompoundModel` instance which is a sum of the given PSF
+            models.
+        """
+
+        psf_class = type(self.psf)
+        sum_psf = psf_class(sigma=self.psf.sigma.value,
```
hamogu: How would one fix certain parameters and make sure they stay fixed for all fits?
Commonly, one would fit the sigma only on the brightest stars and then freeze it in the fit. That has two advantages: 1) The number of fit parameters is lower, makes for a faster fit and 2) sigma really is a property of the telescope and the weather at the time of observations, so it should be the same for all stars. Fits often go bad when there are several faint stars on a background, e.g. one of them gets fit with sigma = 1e5, which is essentially a constant background.
Similarly, one might want to keep x_0 and y_0 fixed. For example, imagine that I already know the positions of the stars from the HIPPARCOS catalog, which is a lot more precise than any ground-based data that I have. In that case, I only want to fit the fluxes.

photutils/psf/psfphotometry.py
mirca: I think I forgot to delete this, haha :)

photutils/psf/psfphotometry.py
mirca: Thanks! I did notice your ``extract_array`` implementation!

hamogu: I'm not sure is `extract_array` itself is the the thing to sue here, but maybe `overlap_slices`...

photutils/psf/psfphotometry.py
bsipocz: Indeed, it should be ``astropy.extern.six``. The logic is that we use the bundled version rather than having another runtime dependency.

Also following PEP8, it should be one package import per line

photutils/psf/psfphotometry.py
bsipocz: You can also include the ``~`` at the beginnig there.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
...
+            and the original image.
+        """
+
+        return self.do_photometry(image)
+
+    def _nstar(self, image, star_groups):
+        """
+        Fit, as appropriate, a compound or single model to a given `groups` of
+        stars. Groups are fitted sequentially from the smallest to the biggest. In
+        each iteration, `image` is subtracted by the previous fitted group.
```
bsipocz: Use double ticks for all the parameter names (``image`` here), otherwise sphinx will fail (when trying to resolve them as links)

photutils/psf/psfphotometry.py
bsipocz: I think we shouldn't have any prints, unless we do it with the logger.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,282 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""Module which provides classes to perform PSF Photometry"""
+
+from __future__ import division
+import abc, six
+import numpy as np
+from astropy.table import Table, vstack
+from photutils.psf import subtract_psf
```
bsipocz: Could you please use relative imports?

mirca: Sure, thanks @bsipocz

photutils/psf/psfphotometry.py
larrybradley: Instead of `"shape"` keywords, lately I've been using `"size"` keywords (e.g. following `scipy.ndimage` filters) where a scalar size is also allowed (in that case the size is broadcast to use a square region).

photutils/psf/psfphotometry.py
larrybradley: Yes, absolutely no prints.  If you **have** to, then use `log.info`.

photutils/psf/psfphotometry.py
larrybradley: I think I would have chosen "DAOPhot" or "DAOPhotPhotometry" since the name DAOPHOT is well known and the intent is clear.  I'm not so sure everyone knows that Stetson wrote it.

bsipocz: I'm pretty sure many people don't know it, e.g. I didn't  when I used
daophot as an undergrad, and only got the gotcha after used his variability
index, too for my master's thesis.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,342 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""Module which provides classes to perform PSF Photometry"""
+
+from __future__ import division
+import abc
+import numpy as np
+from astropy.extern import six
+from astropy.table import Table
+from astropy.table import vstack
```
bsipocz: Sorry, I didn't mean to have these in separate lines. Importing several packages in a line is not advised, while several submodules/functions from the same package is fine.

https://www.python.org/dev/peps/pep-0008/#imports

bsipocz:
@mirca - Since this PR need #369, I made a branch to check how happy/unhappy travis really is:
https://travis-ci.org/bsipocz/photutils/builds/143905284

So it's quite good, but some of the tests still needs to be decorated to be skiped when scipy is not available, and also there are a few sphinx warnings.

Also there seems to be an issue with older numpies, I haven't digged into it, but I can if you like before tomorrow's telecon.

mirca:
Great @bsipocz, thank you!
I'm going to rebase #369 soon

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,342 @@
...
+            tab, residual_image = self.nstar(residual_image, star_groups)
+
+            # mark in which iteration those sources were fitted
+            tab['iter_detected'] = n*np.ones(tab['x_fit'].shape, dtype=np.int)
+
+            # populate output table
+            outtab = vstack([outtab, tab])
+
+            # find remaining sources in the residual image
+            sources = self.find(residual_image)
```
bsipocz: This line should catch the warning: ``WARNING: No sources were found. [photutils.detection.findstars]``, as it's rather confusing to get the warning but obviously having sources detected in previous iterations.

mirca:
@bsipocz Would you happen to know why ``appveyor`` and ``travis`` are not happy with ``numpy.assert_raises``?

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,311 @@
...
+        indices = np.indices(image.shape)
+
+        for n in range(len(star_groups.groups)):
+            group_psf = self._get_sum_psf_model(star_groups.groups[n])
+            y = []
+            x = []
+            data = []
+            for row in star_groups.groups[n]:
+                pos = (row['y_0'], row['x_0'])
+                y = np.hstack((y, extract_array(indices[0], self.fitshape,\
```
hamogu: For Eric's use case - probably the fastest and easiest way:
```
usepixel = np.zeros_like(image)
for row in star_groups.groups[n]:
    usepixel[astropy.nddata.utils.overlap_slices(image.shape, self.fitshape, indices, mode='trim') ] =1
x = x[usepixel]
y = y[usepixel]
data = data[usepixel]
```

hamogu: `np.ravel_multi_index` might also help here.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,311 @@
...
+            with any other and label them as beloging to the same group.
+            ``group`` receives as input an `~astropy.table.Table` object
+            with columns named as ``id``, ``x_0``, ``y_0``, in which ``x_0``
+            and ``y_0`` have the same meaning of ``xcentroid`` and
+            ``ycentroid``. This callable must return an `~astropy.table.Table`
+            with columns ``id``, ``x_0``, ``y_0``, and ``group_id``. The
+            column ``group_id`` should cotain integers starting from ``1``
+            that indicate in which group a given source belongs to.
+            See, e.g., `~photutils.psf.DAOGroup`
+        bkg : callable or instance of any BackgroundBase subclasses
```
larrybradley: replace BackgroundBase with `~photutils.BackgroundBase`.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,311 @@
...
+        self.group = group
+        self.bkg = bkg
+        self.psf = psf
+        self.fitter = fitter
+        self.niters = niters
+        self.fitshape = fitshape
+
+        @property
+        def niters(self):
+            return self._niters
```
larrybradley: Need a blank line before the setter decorator.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,311 @@
...
+        @niters.setter
+        def niters(self, niters):
+            if isinstance(niters, int) and niters > 0:
+                self._niters = niters
+            else:
+                raise ValueError('niters must be an interger-valued number, '
+                                 'received niters = {}'.format(niters))
+        @property
+        def fitshape(self):
+            return self._fitshape
```
larrybradley: Need a blank line before the setter decorator.

mirca: @larrybradley  Gotcha!

photutils/psf/psfphotometry.py
mirca: Thanks!

photutils/psf/psfphotometry.py
mirca: Thanks @hamogu!

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,333 @@
...
+            estimations.
+        residual_image : array-like, `~astropy.io.fits.ImageHDU`,
+        `~astropy.io.fits.HDUList`
+            Residual image calculated by subtracting the fitted sources
+            and the original image.
+        """
+
+        return self.do_photometry(image)
+
+    def do_photometry(self, image):
```
bsipocz: For list-driven photmetry I suggest to have an optional kwarg where the user specifies the positions. If it's given we don't do source detections. @eteq @hamogu: what do you say?

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,378 @@
...
+                                  x_0=self.star_group['x_0'][0],
+                                  y_0=self.star_group['y_0'][0],
+                                  fixed=self.psf.fixed)
+            for i in range(len(self.star_group) - 1):
+                group_psf += psf_class(sigma=self.psf.sigma.value,
+                                       flux=self.star_group['flux_0'][i+1],
+                                       x_0=self.star_group['x_0'][i+1],
+                                       y_0=self.star_group['y_0'][i+1],
+                                       fixed=self.psf.fixed)
+            return group_psf
```
hamogu: `return self.finalize_model(group_psf)`

hamogu: default would be
```
def finanlize_model(self, group_psf):
    return group_psf
```

but a derived class could do:
```
class PSFFixedPositions(GroupPSF):
    def finilize_model(self, group_psf):
          for g in group_psf.submodel  # not sure hot to get those, check astroy docs
               g.parameters['x_0'].fixed = True   # note sure of syntax, have never used astropy.modeling
    return group_psf
```

bsipocz:
For psf photometry I think it's fair to ask to have a recent, at least 1.2 astropy, so no need to figure out a workaround for the LTS branches, just skip the tests and maybe raise an error/warning that never astropy is required.
@eteq @hamogu?

mirca:
@bsipocz Travis failures do not appear locally.

mirca:
@eteq Could you take a look in the way I'm using the aperture photometry interface?

bsipocz:
@mirca - You're using functions that are added e.g. in astropy 1.2, thus travis is failing for the LTS tests. You can just simply skip the tests if astropy is older that 1.2. However a note for the docs, or a warning/exception in the code would be useful, too stating this requirement.

photutils/psf/__init__.py
```diff
@@ -7,3 +7,5 @@
 from .funcs import *
 from .models import *
 from .groupstars import *
+from .psfphotometry import *
```
eteq: maybe rename `psfphotometry` to just `photometry`?  after all, it's in `psf` already, so `psf.photometry` makes more sense than `psf.psfphotometry`.

(both the import here and the file name)

photutils/psf/__init__.py
```diff
@@ -7,3 +7,5 @@
...
+from .groupstars import *
```
eteq: isn't this line repeated just above?

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,446 @@
...
+        self.fitshape = fitshape
+        self.aperture_radius = aperture_radius
+
+        @property
+        def niters(self):
+            return self._niters
+
+        @niters.setter
+        def niters(self, niters):
+            if isinstance(niters, int) and niters > 0:
```
eteq: I don't think you want to use `isinstance` here.  `isinstance` on numeric types tends to have unintended side-effects.  For example, in Python 2 "long integers" are a different type from "regular" integers.  So ``isinstance(2L, int)`` evaluates to False, but it should be reasonable to use as `niters`.  Similarly, if I accidentally pass in a float that has an integer value (e.g. 4.0) it should be acceptable. So instead I'd suggest something like this:

```
def niters(self, niters):
    try:
        niters = int(niters)
    except:
        raise ValueError('niters must be an integer or convertable into an integer')
    if niters<=0:
        raise ValueError('niters must be positive')
    self._niters = niters
```
That way *anything* that can be successfully cast into an int will work.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,446 @@
...
+                raise ValueError('fitshape must have two dimensions, '
+                                 'received fitshape = {}'.format(fitshape))
+
+        @property
+        def aperture_radius(self):
+            return self._aperture_radius
+
+        @niters.setter
+        def aperture_radius(self, aperture_radius):
+            if isinstance(aperture_radius, (int, float)) and aperture_radius > 0:
```
eteq: same here as for `niters`: try to cast to ``float`` and raise an error only if that fails.

eteq: also, doesn't this prevent the "None" value from passing through?  That is, with this check in place, `None` can't be the value of `paerture_radius`?

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,446 @@
...
+                       names=('id', 'group_id', 'x_fit', 'y_fit', 'flux_fit',
+                              'iter_detected'),
+                       dtype=('i4', 'i4', 'f8', 'f8', 'f8', 'i4'))
+
+        residual_image = image.copy()
+        residual_image = residual_image - self.bkg(image)
+        sources = self.find(residual_image)
+        
+        if self.aperture_radius is None:
+            self.aperture_radius = self.psf.sigma.value*gaussian_sigma_to_fwhm
```
eteq: `psf` objects aren't guaranteed to have a `sigma` - gaussians do, but certainly many other types of psfs won't (e.g. empirical psfs).  I suggest something like this:
```

if hasattr(self.psf, 'fwhm'):
    self.aperture_radius = self.psf.fwhm
elif hasattr(self.psf, 'sigma'):
    self.aperture_radius = self.psf.sigma.value*gaussian_sigma_to_fwhm
else:
    raise ValueError('PSF does not have sigma or fwhm, so cannot guess the width for aperture photometry')
```

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,446 @@
...
+            the following relative pixel positions: [-2, -1, 0, 1, 2].
+            Also, each element of ``fitshape`` must be an odd number.
+        fitter : Fitter instance (default=LevMarLSQFitter())
+            Fitter object used to compute the optimized centroid positions
+            and/or flux of the identified sources. See
+            `~astropy.modeling.fitting` for more details on fitters.
+        niters : int (default=3)
+            Number of iterations to perform the loop FIND, GROUP, SUBTRACT,
+            NSTAR.
+        aperture_radius : float (default=None)
```
eteq: I like the fact that the default is to try the fwhm, but it would be nice if one could specify *fractions* of FWHM's.  Maybe instead of using None, you can say that any value of `aperture_radius` less than 0 is a FWHM fraction?  So then the default could just be ``-1.0`` to get the same behavior? (and then my comment below about accounting for the None case can be ignored)

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,446 @@
...
+            The radius used to compute initial estimates for the fluxes of
+            sources. If ``None``, one fwhm will be used.
```
eteq: Need to specify the unit here somewhere.  E.g. "the radius in pixels" instead of just "the radius"

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,446 @@
...
+from astropy.stats import gaussian_sigma_to_fwhm
+from ..psf import subtract_psf
+from ..aperture import CircularAperture, aperture_photometry
+
+
+__all__ = ['PSFPhotometryBase', 'DAOPhotPSFPhotometry']
+
+
+@six.add_metaclass(abc.ABCMeta)
+class PSFPhotometryBase(object):
```
eteq: For now I think you should do away with the base class here.  I think it's not clear that the `do_photometry` function would look the same for other approaches.  Unlike things like the star-finders and groupers, the "main task" here is a bit less clear.  And because this is the "highest level" task, it's less likely to need to have a fromalized interface.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,446 @@
...
+                apertures = CircularAperture((sources['xcentroid'],
+                                              sources['ycentroid']),
+                                             r=self.aperture_radius)
+                sources['flux'] = aperture_photometry(residual_image,
+                                                      apertures)['aperture_sum']
+            n += 1
+        return outtab, residual_image
+
+    def do_fixed_photometry(self, image, positions):
+        """
```
eteq: This needs an intro/description.  I.e., what is "fixed" meaning in this context?

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,446 @@
...
+            and the original image.
+        """
+        
+        if positions is None:
+            return self.do_photometry(image)
+        else:
+            return self.do_fixed_photometry(image, positions)
+
+    def do_photometry(self, image):
+        """
```
eteq: This needs a description.  What makes this different from "fixed" photometry?

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,446 @@
...
+        sources = self.find(residual_image)
+        
+        if self.aperture_radius is None:
+            self.aperture_radius = self.psf.sigma.value*gaussian_sigma_to_fwhm
+
+        apertures = CircularAperture((sources['xcentroid'],
+                                      sources['ycentroid']),
+                                     r=self.aperture_radius)
+
+        sources['flux'] = aperture_photometry(residual_image,
```
eteq: very minor, but you might call this "aperture_flux" or similar to distinguish from the psf-fitted flux.  this table is only used internally so that's not a big deal, but it might be useful for debugging or others who want to hack on this in the future or the like.

photutils/psf/psfphotometry.py
```diff
@@ -0,0 +1,446 @@
...
+        star_groups = self.group(intab)
+        outtab, residual_image = self.nstar(residual_image, star_groups)
+        
+        return outtab, residual_image
+
+    def get_uncertainties(self):
+        """
+        Return the uncertainties on the fitted parameters
+        """
+        pass
```
eteq: perhaps ``raise NotImplementedError`` for now, instead of just passing?

eteq:
@mirca - OK, I did a line-by-line review of this, and see my comments above.  Nothing too major, so this is looking good big-picture! The aperture photometry stuff looks good - that's just what I had in mind.

One thing I don't get though: why are `do_photometry` and `do_fixed_photometry` completely separated?  It may be that I'm mis-understanding what you mean by "fixed" photometry, but it looks to me look the main difference is whether or not the user provides an initial list of stars or if they let the finder do it instead.  In my experience "fixed" photometry generally instead means that the *position* is fixed:  e.g., the ``x`` and ``y`` parameters of the psf model are set to fixed so that the fitter *only* fits the flux(es).  In that case the implementation should be basically the same: it's just whether or not the parameters are fixed on the psf model that controls the "fixed" behavior.  (if it's easier, feel free to plan to just discuss this at our next telecon)

larrybradley:
To add on @eteq's comment, what you have named "fixed" photometry here is generally called "forced" photometry.

bsipocz:
> To add on @eteq's comment, what you have named "fixed" photometry here is generally called "forced" photometry.

(or list driven photometry).

larrybradley:
@mirca I would suggest combining `do_photometry` and `do_fixed_photometry` as @eteq suggested.  If that is not possible for some reason, I would suggest `do_forced_photometry` or if you want to be descriptive, `do_fixed_position_photometry`.

bsipocz:
If I would have a vote, I would go with Erik's suggestion to have the positions as an optional kwarg in ``do_photometry``. If it stays as a separate method, go either forced or fixed as list driven is unnecessary long.

photutils/psf/photometry.py
```diff
@@ -0,0 +1,426 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""Module which provides classes to perform PSF Photometry"""
+
+from __future__ import division
+import abc
```
hamogu: Is `abc` actually used anywhere?

photutils/psf/photometry.py
```diff
@@ -0,0 +1,426 @@
...
+    SUBTRACT, FIND until no more stars are detected or a given number of
+    iterations is reached.
+    """
+
+    def __init__(self, group, bkg, psf, fitshape, find=None,
+                 fitter=LevMarLSQFitter(), niters=3, aperture_radius=None):
+        """
+        Attributes
+        ----------
+        find : callable or instance of any StarFinderBase subclasses
```
hamogu: Attributes should be documented in the same order they appear in the function signature.

photutils/psf/photometry.py
```diff
@@ -0,0 +1,426 @@
...
+            ``find`` should be able to identify stars, i.e. compute a rough
+            estimate of the centroids, in a given 2D image.
+            ``find`` receives as input a 2D image an return an
+            `~astropy.table.Table` object which contains columns with names:
+            ``id``, ``xcentroid``, ``ycentroid``, and ``flux``. In which
+            ``id`` is an interger-valued column starting from ``1``,
+            ``xcentroid`` and ``ycentroid`` are center position estimates of
+            the sources and ``flux`` contains flux estimates of the sources.
+            See, e.g., `~photutils.detection.DAOStarFinder`
+        group : callable or instance of any GroupStarsBase subclasses
```
hamogu: Set such that GroupStarBase is found and linked with sphinx in the same wa you di dthat fir e.g. DAOStarFinder in the line just above.

photutils/psf/photometry.py
```diff
@@ -0,0 +1,426 @@
...
+            ``group`` should be able to decide whether a given star overlaps
+            with any other and label them as beloging to the same group.
+            ``group`` receives as input an `~astropy.table.Table` object
+            with columns named as ``id``, ``x_0``, ``y_0``, in which ``x_0``
+            and ``y_0`` have the same meaning of ``xcentroid`` and
+            ``ycentroid``. This callable must return an `~astropy.table.Table`
+            with columns ``id``, ``x_0``, ``y_0``, and ``group_id``. The
+            column ``group_id`` should cotain integers starting from ``1``
+            that indicate in which group a given source belongs to.
```
hamogu: remove "in"

photutils/psf/photometry.py
```diff
@@ -0,0 +1,426 @@
...
+            See, e.g., `~photutils.psf.DAOGroup`
+        bkg : callable or instance of any `~photutils.BackgroundBase`
+            subclasses ``bkg`` should be able to compute either a scalar
```
hamogu: I think the line break should be after "subclass" not before it.

photutils/psf/photometry.py
```diff
@@ -0,0 +1,427 @@
...
+            ``xcentroid`` and ``ycentroid`` are center position estimates of
+            the sources and ``flux`` contains flux estimates of the sources.
+            See, e.g., `~photutils.detection.DAOStarFinder`
+        group : callable or instance of any GroupStarsBase subclasses
+            ``group`` should be able to decide whether a given star overlaps
+            with any other and label them as beloging to the same group.
+            ``group`` receives as input an `~astropy.table.Table` object
+            with columns named as ``id``, ``x_0``, ``y_0``, in which ``x_0``
+            and ``y_0`` have the same meaning of ``xcentroid`` and
+            ``ycentroid``. This callable must return an `~astropy.table.Table`
```
hamogu: It's acutally more than a bit annoying that the same thing is sometimes called _centroid and  sometimes _0. I know that that naming is used in the finders. If we cannot unify it in this PR, can you open an issue so that we come back to it after this is merged?

photutils/psf/photometry.py
```diff
@@ -0,0 +1,427 @@
...
+            to suitable starting values for each fit. The names of these
+            parameters should be given as ``x_0``, ``y_0`` and ``flux``.
+
+            `~photutils.psf.prepare_psf_model` can be used to prepare any 2D
+            model to match this assumption.
+        fitshape : array-like
+            Rectangular shape around the center of a star which will be used
+            to collect the data to do the fitting, e.g. (5, 5) means to take
+            the following relative pixel positions: [-2, -1, 0, 1, 2].
+            Also, each element of ``fitshape`` must be an odd number.
```
hamogu: Why "Also"?
Also, maybe array-like could also say that it's 2-d.

photutils/psf/photometry.py
```diff
@@ -0,0 +1,427 @@
...
+        fitter : Fitter instance (default=LevMarLSQFitter())
+            Fitter object used to compute the optimized centroid positions
+            and/or flux of the identified sources. See
+            `~astropy.modeling.fitting` for more details on fitters.
+        niters : int (default=3)
```
hamogu: No need to give defaults here. Sphinx shows the function signature which contains the default. If we ever change the default, we will forget to change it here so better not list is here.

photutils/psf/photometry.py
```diff
@@ -0,0 +1,427 @@
...
+            Number of iterations to perform the loop FIND, GROUP, SUBTRACT,
+            NSTAR.
+        aperture_radius : float (default=None)
```
hamogu: No default here either.

photutils/psf/photometry.py
```diff
@@ -0,0 +1,427 @@
...
+            The radius (in units of pixels) used to compute initial estimates
+            for the fluxes of sources. If ``None``, one fwhm will be used.
```
hamogu: Mmm. What happens if there is no FWHM? Gaussian PSFs have a FWHM, but `psf` could be an arbitrary model that does not have this parameter.

photutils/psf/photometry.py
bsipocz: On the other hand we always list the defaults too, but not in this first line of the given parameter, but at the end of the description below.

photutils/psf/photometry.py
```diff
@@ -0,0 +1,427 @@
...
+
+    def do_photometry(self, image, positions=None):
+        """
+        Perform PSF photometry in ``image``. This method assumes that
+        ``psf`` has centroids and flux parameters which will be fitted to the
+        data provided in ``image``. A compound model, in fact a sum of
+        ``psf``, will be fitted to groups of starts automatically identified
+        by ``group``. Also, ``image`` is not assumed to be background
+        subtracted.
+        If positions are not ``None`` then this method performs forced PSF
```
hamogu: I know we discussed it already, but I don't think that "forced PSF photometry" is a well-known term, e.g. I don't know what exactly that means. So there should e an extra sentence here to explain what is "forced"

photutils/psf/photometry.py
```diff
@@ -0,0 +1,427 @@
...
+        as ``fixed`` in the PSF model ``psf``, then the optimizer will only
+        consider the flux as a variable. Otherwise, ``positions`` will be used
+        as initial guesses for the centroids.
+
+        Parameters
+        ----------
+        image : 2D array-like, `~astropy.io.fits.ImageHDU`,
+        `~astropy.io.fits.HDUList`
+            Image to perform photometry.
+        positions: `~astropy.table.Table` (default=None)
```
hamogu: No default here...

photutils/psf/photometry.py
```diff
@@ -0,0 +1,427 @@
...
+            each object. Columns 'x_0' and 'y_0' must be present.
+            'flux_0' can also be provided to set initial fluxes.
+        
+        Returns
+        -------
+        outtab : `~astropy.table.Table`
+            Table with the photometry results, i.e., centroids and fluxes
+            estimations.
+        residual_image : array-like, `~astropy.io.fits.ImageHDU`,
+        `~astropy.io.fits.HDUList`
```
hamogu: I don't think the line brake between 216/217 will work without confusing sphinx

photutils/psf/photometry.py
```diff
@@ -0,0 +1,427 @@
...
+                             'number, received aperture_radius = {}'
+                             .format(value))
+
+
+    def __call__(self, image, positions=None):
+        """
+        Parameters
+        ----------
+        image : 2D array-like, `~astropy.io.fits.ImageHDU`,
+        `~astropy.io.fits.HDUList`
```
hamogu: line break correct?

photutils/psf/photometry.py
```diff
@@ -0,0 +1,427 @@
...
+        def __init__(self, psf, star_group):
+            self.star_group = star_group
+            self.psf = psf
+        
+        def get_model(self):
+            """        
+            Returns
+            -------
+            group_psf : CompoundModel
+                `CompoundModel` instance which is a sum of the given PSF
```
hamogu: I really did read to the end, but I don't have comments here.

photutils/psf/tests/test_photometry.py
```diff
@@ -0,0 +1,202 @@
...
+        assert_array_equal(result_tab['x_fit'], sources['x_mean'])
+        assert_array_equal(result_tab['y_fit'], sources['y_mean'])
+        assert_allclose(result_tab['flux_fit'], sources['flux'], rtol=1e-1)
+        assert_array_equal(result_tab['id'], sources['id'])
+        assert_array_equal(result_tab['group_id'], sources['group_id'])
+        assert_allclose(np.mean(residual_image), 0.0, atol=1e1)
+
+
+    def test_complete_photometry_two(self):
+        sigma_psf = 2.0
```
hamogu: short sentence what's different from test one. Just if this fails and somebody needs to understand what's going on five years from now (or, looking at how long IRAF is already used in astronomy, 30 years from now when we try to run this on Python 7.1)

mirca:
@eteq if you are happy with the changes, I think you can merge this =) Also, please take a look at this notebook [here](https://github.com/astropy/photutils-datasets/pull/5)

mirca:
@larrybradley Yes, I incorrectly rebased, thanks! From the "Files changed tab", I think it's fine now.

larrybradley:
I'd like to review this before it's merged (if it's near complete).  In the meantime, @mirca can you please squash some of these commits -- in particular I see several back-to-back commits with the same message (e.g. "add aperture radius parameter", "add peakfinder tests", "spliting in different PRs", etc.).

mirca:
@larrybradley Sure! I would appreciate if you could take a look on this too!

mirca:
@eteq Not sure why Travis is complaining about ``AttributeError: x_0``. I don't see this locally.

larrybradley:
@mirca It's still not correct.  You'll need to pull the current master and rebase onto that.  There may be merge conflicts, which will need to be resolved (i.e. that last commit, which I don't think should be there).  Once reconciled, it will then require a forced push of this PR.

larrybradley:
It does seem right that Github is reporting no conflicts, because I think there are with `findstars.py`.

mirca:
@larrybradley I think I did rebase correctly this time =D

photutils/detection/findstars.py
```diff
@@ -326,61 +326,6 @@ def __init__(self, threshold, fwhm, sigma_radius=1.5, minsep_fwhm=2.5,
         self.exclude_border = exclude_border
 
     def find_stars(self, data):
-        """
```
larrybradley: I'm still confused about this commit.  The lines that you remove here (329-383) are *not* in the current master version of `findstars.py` (i.e. I already removed them):
https://github.com/astropy/photutils/blob/master/photutils/detection/findstars.py#L329

bsipocz:
> Not sure why Travis is complaining about AttributeError: x_0. I don't see this locally.

Have you tried it locally with old numpies? There must be something different in np<1.9. Having a very quick look I don't see anything obvious in astropy modeling, but it would be great if you could find the solution. Otherwise I'm also OK for having to require np>=1.9 for photutils.psf (but then it should be stated in the docs, and possibly raise an error at import time).

It looks like some of the pytest.skip decorators got lost during your rebases. At least the LTS tests used to pass, but now complaining again for the missing ``NoOverlapError``: https://travis-ci.org/astropy/photutils/jobs/153385289

Same problem with scipy in the test ``TestDAOPhotPSFPhotometryAttributes``, it should be skipped if scipy is not available.

mirca:
@bsipocz failures related with scipy and astropy still shows up, although I placed ``@pytest.mark.skipif('not HAS_SCIPY or HAS_MIN_ASTROPY')`` on the top of each class. Locally, the tests are skipped if, for example, I have astropy 1.1

bsipocz:
@mirca - It fails with the imports. You could do that in a try except, and raise the ImportError stating the min. version number required. There are a few examples in astropy, e.g. astropy/visualization/mpl_normalize.py

https://travis-ci.org/astropy/photutils/jobs/153785322#L714

photutils/psf/photometry.py
```diff
@@ -0,0 +1,448 @@
...
+import numpy as np
+from astropy.table import Table, vstack, hstack
+from astropy.modeling.fitting import LevMarLSQFitter
+from astropy.nddata.utils import overlap_slices
+from astropy.stats import gaussian_sigma_to_fwhm
+from ..psf import subtract_psf
+from ..aperture import CircularAperture, aperture_photometry
+
+try:
+    from astropy.nddata.utils import NoOverlapError
```
bsipocz: We've just talked about it, move this import to the place it's used, so the importerror is not coming up at import time with older astropies. Maybe leave a note as well, that this can be cleaned up once stop supporting 1.0.x LTS

bsipocz:
Bad news: the tests still fail with the newer scipy (and in addition a few more failures are coming in, too, e.g. nanmedian is already been removed from the newer scipy, but not yet added in those old numpies).

mirca:
@bsipocz I'm gonna dig into numpy's issue track to see if this requires an upstream fix

bsipocz:
@mirca - Also, maybe you can raise this as an astropy issue with modeling. Many of the tests are skipped when no scipy is installed, and we only test for old numpies without scipy. Maybe someone saw it before or knows which new numpy feature are being used.

photutils/psf/photometry.py
```diff
@@ -355,8 +355,8 @@ def nstar(self, image, star_groups):
             try:
                 from astropy.nddata.utils import NoOverlapError
             except ImportError:
-                raise ImportError("astropy 1.2.1 is required in order to use this
-                                   class.")
+                raise ImportError("astropy 1.2.1 is required in order to use" +
```
bsipocz: super nitpicking: if a string is within parenthesis, you don't need the ``+``, just end the line with the quote mark, and start the new line with it (just as you did).

mirca:
@bsipocz @eteq @hamogu and voila =D the bug occurred when I was using ``np.size(fit_model)`` to compute the number of models in a ``CompoundModel``. For ``numpy >=1.9``, ``np.size(fit_model)`` returns the number of components in a ``CompoundModel`` instance, while for ``numpy < 1.9`` it always returns ``1``. Now, I'm checking if ``fit_model`` has ``submodel_names`` as parameter (which is ``true`` for any ``CompoundModel`` instance) and ``false`` for a "single" ``Model``.

P.S.: @hamogu Since you always can see simple solutions, do you see another way to do that?

bsipocz:
This is great investigation, thanks @mirca.

Note that travis currently has issues with all the OSX builds, so I consider this all green now.

larrybradley:
@mirca I think you want `hasattr(fit_model, submodel_names)`.

hamogu:
I think that's fine as it is now. I don't see a simpler way.

mirca:
@larrybradley Not sure if I understood your request. I think I'm using ``hasattr(fit_model, 'submodel_names')``

larrybradley:
@mirca Ignore my comment -- I didn't look at your code.  Sorry for the noise.

mirca:
@eteq Do you think this is mergeable?

hamogu:
@mirca: Can you squash the commits a little? It's fine to have one than one commit per PR (say, maybe 10 or so), but this are a lot and in some cases the commit messages indicate that consecutive commits address the same thing (e.g. "docs" and "most docs"). A smaller number of commits makes it easier for someone who later needs to see how things changed. For that, you don't need e.g. "fix indentation". You could merge that with the previous commit and just pretend that you typed it in correctly in the first place.
Do you know how to interactively squash commits? If not, let me know. 
The most important advice is probably to make a copy of your branch before you do anything, just in case.

photutils/psf/photometry.py
```diff
@@ -0,0 +1,450 @@
...
+    (1987) to perform point spread function photometry in crowded fields,
+    which consists basically in applying the loop FIND, GROUP, NSTAR,
+    SUBTRACT, FIND until no more stars are detected or a given number of
+    iterations is reached.
+    """
+
+    def __init__(self, group, bkg, psf, fitshape, find=None,
+                 fitter=LevMarLSQFitter(), niters=3, aperture_radius=None):
+        """
+        Attributes
```
larrybradley: Attributes -> Parameters

photutils/psf/photometry.py
```diff
@@ -0,0 +1,450 @@
...
+        ----------
+        group : callable or instance of any `~photutils.psf.GroupStarsBase`
+            subclasses ``group`` should be able to decide whether a given star
```
larrybradley: "subclasses" needs to go at the end of the previous line.  Unfortunately, the parameter/type line cannot be wrapped.  It's a case where PEP8 line-length can be ignored since there is no alternative (that I'm aware of).

photutils/psf/photometry.py
larrybradley: This needs to be on one line.

photutils/psf/photometry.py
larrybradley: Yep, here too.

photutils/psf/photometry.py
```diff
@@ -0,0 +1,450 @@
...
+        photometry, i.e., the positions are assumed to be known with high
+        accuracy and only fluxes are fitted. If the centroid positions are
+        set as ``fixed`` in the PSF model ``psf``, then the optimizer will only
+        consider the flux as a variable. Otherwise, ``positions`` will be used
+        as initial guesses for the centroids.
+
+        Parameters
+        ----------
+        image : 2D array-like, `~astropy.io.fits.ImageHDU`,
+        `~astropy.io.fits.HDUList`
```
larrybradley: 1 line here too.

photutils/psf/photometry.py
```diff
@@ -0,0 +1,450 @@
...
+            'flux_0' can also be provided to set initial fluxes.
+        
+        Returns
+        -------
+        outtab : `~astropy.table.Table`
+            Table with the photometry results, i.e., centroids and fluxes
+            estimations and the initial estimates used to start the fitting
+            process.
+        residual_image : array-like, `~astropy.io.fits.ImageHDU`,
+        `~astropy.io.fits.HDUList`
```
larrybradley: And here...

photutils/psf/photometry.py
```diff
@@ -0,0 +1,450 @@
...
+            Residual image calculated by subtracting the fitted sources
+            and the original image.
+        """
+        
+        residual_image = image.copy()
+        residual_image = residual_image - self.bkg(image)
```
larrybradley: You could replace these two lines with simply:
```residual_image = image - self.bkg(image)```

File: photutils/psf/photometry.py [hunk 69]
```diff
@@ -0,0 +1,450 @@
...
+                outtab = vstack([outtab, result_tab])
+
+                sources = self.find(residual_image)
+
+                if len(sources) > 0:
+                    apertures = CircularAperture((sources['xcentroid'],
+                                                  sources['ycentroid']),
+                                                 r=self.aperture_radius)
+                    sources['aperture_flux'] = aperture_photometry(residual_image,\
+                                                apertures)['aperture_sum']
```
larrybradley: The "\" is unnecessary in line 274 and line 275 is not properly indented.

photutils/psf/photometry.py
```diff
@@ -0,0 +1,450 @@
...
+
+            outtab = hstack([intab, outtab])
+        else:
+            if 'flux_0' not in positions.colnames:
+                apertures = CircularAperture((positions['x_0'],
+                                              positions['y_0']),
+                                             r=self.aperture_radius)
+
+                positions['flux_0'] = aperture_photometry(residual_image,\
+                                             apertures)['aperture_sum']
```
larrybradley: Again, no "\" and fix the indentation of the second line.

mirca:
@hamogu Sure!

photutils/psf/photometry.py
```diff
@@ -0,0 +1,450 @@
...
+                                  fixed=self.psf.fixed, tied=self.psf.tied,
+                                  bounds=self.psf.bounds)
+            for i in range(len(self.star_group) - 1):
+                group_psf += psf_class(sigma=self.psf.sigma.value,
+                                       flux=self.star_group['flux_0'][i+1],
+                                       x_0=self.star_group['x_0'][i+1],
+                                       y_0=self.star_group['y_0'][i+1],
+                                       fixed=self.psf.fixed,
+                                       tied=self.psf.tied,
+                                       bounds=self.psf.bounds)
```
larrybradley: Why is the `group_psf` assignment split here?  Can't you simply do:

```
group_psf = []
for i in range(len(self.star_group) - 1):
    group_psf.append(psf_class(...... [i]))    # index i on flux, x_0, y_0
```

larrybradley: Nevermind, please ignore.  I see that you are making a CompoundModel.  Sorry.

hamogu:
@mirca : While all of @larrybradley 's comments makes sense, I should point out that the mere fact that you get comments on this level means that this PR could be merged right now.

mirca:
@hamogu Don't merge it right now please, I'm just squashing the commits =D

mirca:
@larrybradley Thanks for the comments! :D

hamogu:
Don't worry. I don't have the right to merge anything in any astropy
project. I just wanted to make sure you don't get discouranged the the
sheer number of comments coming in.

larrybradley:
@hamogu  Yes, it looks mostly fine (barring some minor comments).  However, I didn't look too deeply at the algorithm and implementation details.  I assumed that's already been worked out.  :smile:

Thanks, @mirca!  I hope I didn't overwhelm you with minor comments.

mirca:
@hamogu @larrybradley Maybe better squash all of the commits into one, just with the commit message "add daophot-based PSF photometry for crowded fields"?

larrybradley:
@mirca I think squashing into only one commit is too extreme.  What you have now with 14 commits looks good to me.

hamogu:
@mirca You did a lot of work on this and that can show in 14 commits. It's not about squashing everything but in striking a balance between having enough commits to show how it was developed but having to few commits that it's still possible to read the commit messages in a reasonable time.
(There are some projects that always squash and use a "1PR = 1 commit" rule, but astropy is not one of them.)

photutils/psf/photometry.py
eteq: removed the also in merging.  Thought about also saying it's length-2, but I realized I wanted to make a slight change in that anyway, so can address this in the PR I'm about to submit

photutils/psf/photometry.py
eteq: fixed while merging

photutils/psf/photometry.py
eteq: fixed while merging

eteq:
@hamogu @larrybradley @mirca - I fully agree squashing into one is not necessary - this is good because it actually captures the logical development, and you never know when that might be useful!  I tend to have a "the more commits the better" attitude, although that's to a certain extent a matter of taste.  I agree that two commits along the lines of "docs" and "more docs" should be squashed, but it's good if commits come in reasonably logical "chunks" in case you ever need to sniff out where a bug got introduced.

At any rate, I had another look over this.  I think it's great aside from a couple very minor things that seemed to slip through, and some more substantive naming concerns I had... *but* the latter set don't affect the functionality, so here's what I'll do: I'm going to merge this now (after I manually add a commit fixing a few of the minor things), and then I'll create my own PR with the larger changes.  That way this work gets in and we can have a separate conversation about the naming.  Thanks for all the work, @mirca!

bsipocz:
Hurray, this is merged :) Thank you for all your work @mirca! 🎉

mirca:
Thank you @bsipocz @eteq @hamogu 🎉

PR #392, opened 2016-07-14
Add axis keyword to Background classes

larrybradley:
This will also be used in an object-oriented interface for the 2D background classes.

### DIFF
--- a/photutils/background/core.py
+++ b/photutils/background/core.py
@@ -15,8 +15,8 @@ import warnings
 from astropy.extern import six
 from astropy.utils.misc import InheritDocstrings
 from astropy.utils.exceptions import AstropyUserWarning
-from astropy.stats import sigma_clip, mad_std
-from ..extern.biweight_stats import biweight_location, biweight_midvariance
+from astropy.stats import sigma_clip
+from ..extern.stats import mad_std, biweight_location, biweight_midvariance
 
 import astropy
 if LooseVersion(astropy.__version__) < LooseVersion('1.1'):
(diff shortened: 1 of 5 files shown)


photutils/background/core.py
```diff
@@ -32,6 +32,38 @@
...
 
 
+def _masked_median(data, axis=None):
+    """
+    Calculate the median of a (masked) array.
+
+    This function is necessary for a consistent interface across all
+    numpy versions.  An bug was introduced in numpy v1.10 where
+    `numpy.ma.median` (with ``axis=None``) returns a single-valued
+    ~numpy.ma.MaskedArray` if the input data is a `~numpy.ndarray` or if
```
mirca: ` before ~numpy.ma.MaskedArray

photutils/background/core.py
```diff
@@ -32,6 +32,38 @@
            'MADStdBackgroundRMS', 'BiweightMidvarianceBackgroundRMS']
 
 
+def _masked_median(data, axis=None):
+    """
+    Calculate the median of a (masked) array.
+
+    This function is necessary for a consistent interface across all
+    numpy versions.  An bug was introduced in numpy v1.10 where
+    `numpy.ma.median` (with ``axis=None``) returns a single-valued
```
mirca: ~ here?

larrybradley: No, I wanted to distinguish it as np.ma.median, otherwise it would render simply as "median" (which would be likely interpreted as np.median).  Also, this is a private function, so the the docs will never be shown. :smile:

photutils/background/core.py
larrybradley: Fixed, thanks.

photutils/background/core.py
mirca: xD

PR #397, opened 2016-08-05
checkimage_name option for detect_sources

johnnygreco:
Hello, 

I've added an option (checkimage_name) to save the filtered image generated by detect_sources. This is an option that SExtractor offers, which I find very useful as a sanity check. 

I think my changes are simple and self-explanatory, but this is my first pull request, so please let me know if should have done something differently.

Thanks, 
Johnny

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,6 +33,9 @@ New Features
 
   - Added new ``make_source_mask`` convenience function. [#355]
 
+  - Added ``checkimage_name`` option to ``detect_sources``, which 
+    gives to option to save the filtered image for inspection. 
+
 - ``photutils.psf``
 
   - Added the ``param_uncert`` keyword to ``psf_photometry`` so that
--- a/photutils/detection/core.py
+++ b/photutils/detection/core.py
@@ -135,7 +135,7 @@ def detect_threshold(data, snr, background=None, error=None, mask=None,
 
 
 def detect_sources(data, threshold, npixels, filter_kernel=None,
-                   connectivity=8):
+                   connectivity=8, checkimage_name=None):
     """
     Detect sources above a specified threshold value in an image and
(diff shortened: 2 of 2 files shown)


larrybradley:
@johnnygreco Thanks for your submission.   I understand you want to get the filtered image, but I don't think writing it to a FITS file is the correct approach here.  It would be much better if the filtered array is simply returned from `detect_sources` along with the segmentation image .  However, that would break the API somewhat, so perhaps the best approach is to simply create a new function (e.g. `filter_image`) that generates only the filtered image (this new function would be called internally by `detect_sources`).  Those changes wouldn't affect the `detect_sources` API.  The minor downside is that if you call `filter_image` independently then the convolution must be calculated again, but that's not very time-consuming (and I'm not sure that it'll be frequently used).

johnnygreco:
@larrybradley Thanks for your fast response. I definitely see your point. Isn't `utils.convolution._convolve_data` basically the `filter_image` function you describe? Are you suggesting a `filter_image` function within `detection.core` that calls `_convolve_data`? Or perhaps making `_convolve_data` public?

In any case, I suppose it isn't too difficult for the user to simply convolve the image using `ndimage.convolve`, but it would be nice to have a convenience function within `photutils` that does this in the exact same way as `detect_sources`.  

If you see a path forward that would be useful, I'd be happy to implement it, but I also understand if this seems unnecessary.

larrybradley:
`_convolve_data` is a (private) convenience function to allow for both `np.ndarray` and `astropy.convolution.Kernel2D` kernel inputs.  Essentially what I'm proposing is to make
`_convolve_data` a public function, renaming it to `filter_data`.

johnnygreco:
Okay, thanks. I'll open a new PR, since this is different from my initial submission.

PR #411, opened 2016-09-01
fix see also section that was missing "also"

eteq:
This fixes a warning I've been seeing pop up about an improperly-named section.  (for some reason sphinx doesn't recognize this particular type of "warning" as a sphinx build warning... but it still appears in the logs.  Anyway, this fixes it.)


eteq:
Simple enough, so I'm merging now.  (I think @mirca is the author of this docstring originally, so you might want to just be aware that this section name isn't a valid one for our docs machinery)

PR #418, opened 2016-09-08
Change GroupPSF class to a generator function

### LINKED ISSUE #417: GroupPSF class defined within the DAOPhotPSFPhotometry class
I assume this wasn't intentional?
https://github.com/astropy/photutils/blob/master/photutils/psf/photometry.py#L403

Should `GroupPSF` belong in the `groupstars.py` module?

eteq:
This follows some of the discussion from #417 (that was about whether or not `GroupPSF` should be inside `DAOPhotPSFPhotometry`).  After looking at that I realized not only does it make sense for `GroupPSF` to not be only available from `DAOPhotPSFPhotometry`, but it probably also should be a generator function.

My reasoning is that the `GroupPSF` class is a bit surprising: it *sounds* like it's a model, but it's actually a class with a method that *generates* a model.  I think it's cleaner (and more likely to be re-usable by others) to just change this to a function which returns the new model.  That way there's not confusion between the class that *is* the model and the class that *makes* the model (because after this PR the latter is not a class at all).

This also closes #417 by moving the functionality to the `psf/models.py` module.

cc @larrybradley @mirca

### DIFF
--- a/photutils/psf/models.py
+++ b/photutils/psf/models.py
@@ -574,10 +574,12 @@ def get_grouped_psf_model(template_psf_model, star_group):
 
     Parameters
     ----------
+    template_psf_model : `astropy.modeling.Fittable2DModel` instance
+        The model to use for *individual* objects.  Must have parameters named
+        `x_0`, `y_0`, and `flux`.
     star_group : `~astropy.table.Table`
         Table of stars for which the compound PSF will be constructed.  It
         must have columns named `x_0`, `y_0`, and `flux_0`.
-    template_psf_model : `astropy.modeling.Fittable2DModel` instance
 
     Returns
     -------
--- a/photutils/psf/models.py
+++ b/photutils/psf/models.py
@@ -13,7 +13,8 @@ from ..utils import mask_to_mirrored_num
 from ..extern.nddata_compat import extract_array
 
 
-__all__ = ['IntegratedGaussianPRF', 'PRFAdapter', 'prepare_psf_model']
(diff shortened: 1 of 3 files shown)


photutils/psf/models.py
```diff
@@ -564,3 +565,38 @@ def prepare_psf_model(psfmodel, xname=None, yname=None, fluxname=None,
...
+    """
+    Construct a joint PSF model which consists of a sum of PSF's templated on
+    a specific model, but whose parameters are given by a table of objects.
+
+    Parameters
+    ----------
+    star_group : `~astropy.table.Table`
+        Table of stars for which the compound PSF will be constructed.  It
+        must have columns named `x_0`, `y_0`, and `flux_0`.
+    template_psf_model : `astropy.modeling.Fittable2DModel` instance
```
larrybradley: The parameter order here doesn't match the function signature order.

larrybradley:
Thanks, @eteq.  Looks good, except for my comment about the parameter order in the docs.

photutils/psf/models.py
eteq: good catch, thanks!

eteq:
Oops, there was an issue in the docs.  New commits should do the job (and also added a test).  But now this should be set to merge as soon as the tests finish (assuming they pass).

eteq:
Arg, one more small fix - was missing the `HAS_SCIPY` decorator (didn't realize that was necessary for creating an IntegratedGaussianPRF.  Fixed now, and everything else about the tests worked just fine, so I'll merge this now.  @mirca - feel free to start working on the refactoring now!

PR #420, opened 2016-09-08
Add astropy sigma clipping functions to extern

larrybradley:
The function signature and feature options for `astropy.stats.sigma_clip` changed in `astropy 1.1`.  Instead of adding boilerplate astropy version checking code when `sigma_clip` or `sigma_clipped_stats` is called in `photutils`, I've simply added the latest `astropy` sigma clipping functions in `extern`.  We can remove this and update the `sigma_clip` and `sigma_clipped_stats` imports when we no longer support `astropy 1.0`.

I've updated this in `detection.core` and `background.core`.  I'll also update it in `background.background2d` within a separate PR on 2D backgorunds that I'm currently working on.

### DIFF
--- a/photutils/extern/sigma_clipping.py
+++ b/photutils/extern/sigma_clipping.py
@@ -1,0 +1,296 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+import numpy as np
+import warnings
+from astropy.utils.exceptions import AstropyUserWarning
+
+
+__all__ = ['sigma_clip', 'sigma_clipped_stats']
+
+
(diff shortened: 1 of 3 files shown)


photutils/extern/sigma_clipping.py
```diff
@@ -0,0 +1,296 @@
...
+        which may not be what one wants (especially if filtering was
+        done along an axis).
+
+    Examples
+    --------
+    This example generates random variates from a Gaussian distribution
+    and returns a masked array in which all points that are more than 2
+    sample standard deviations from the median are masked::
+
+        >>> from astropy.stats import sigma_clip
```
mirca: @larrybradley I think you have to update the doctests too

larrybradley: Good catch, thanks!

PR #423, opened 2016-09-12
Make SigmaClip a general class

larrybradley:
With this PR `SigmaClip` is no longer a mixin class.  Instead one must input a `SigmaClip` object to the background classes.

The `SigmaClip` class can be used in any function that performs sigma clipping as part of its functionality.  Instead of passing all of the many `sigma_clip` keywords to such a function, one can simply create a `SigmaClip` object and pass that to any function/class as needed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,8 @@ New Features
 
   - Added new ``std_blocksum`` function. [#355]
 
+  - Added new ``SigmaClip`` class. [#423]
+
 - ``photutils.datasets``
 
   - Added ``load_irac_psf`` function. [#403]
--- a/photutils/background/core.py
+++ b/photutils/background/core.py
@@ -101,7 +101,7 @@ class SigmaClip(object):
         Defaults to the standard deviation (`numpy.std`).
     """
 
-    def __init__(self, sigma=3, sigma_lower=None, sigma_upper=None, iters=5,
+    def __init__(self, sigma=3., sigma_lower=None, sigma_upper=None, iters=5,
                  cenfunc=np.ma.median, stdfunc=np.std):
         self.sigma = sigma
         self.sigma_lower = sigma_lower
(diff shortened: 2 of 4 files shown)


mgennaro:
Hi Larry,
just a quick comment on this, from a user's perspective.
While some may be familiar with object oriented programming, I have a concern that "old fashioned" users may not find it always straightforward to follow this path.
It may also look like too much having to define a background object, then a sigma-clip object, then a group object, then a fitter, a psf, and so on (I love the idea, but I am being the devil's advocate here).
Are there going to be " default" settings when initializing the background class? Something that initializes the sigma-clipper to a typical 3sigma from the median, 5 iters thing? Will it be possible for the user to access the default object, then have a look at it, and maybe learn how to modify its parameters by trial and error?

larrybradley:
@mgennaro Yes, the default for the background objects has `sigma=3` and `iters=5`, see here:  https://github.com/astropy/photutils/blob/master/photutils/background/core.py#L159

Hopefully this isn't too much object-oriented programming.  The background classes all have examples that show how it's used:  http://photutils.readthedocs.io/en/latest/api/photutils.background.MMMBackground.html#photutils.background.MMMBackground

I also plan to add some narrative docs for the new background classes.

As a simple example, instead of:
```
bkg = MedianBackground(sigma=3, iters=5)
```

One would do:
```
bkg = MedianBackground(SigmaClip(sigma=3, iters=5))
```

or

```
bkg = MMMBackground(sigma_clip=SigmaClip(sigma=3, iters=5))
```


One advantage of this is that the `SigmaClip` class has the full suite of `sigma_clip` options (e.g. `sigma_lower`, `sigma_upper`, `cenfunc`, `stdfunc`) whereas previously the background classes did not.

larrybradley:
I should add that using the defaults (`sigma=3` and `iters=5`), one can simply do:

```
bkg = MedianBackground()
```

And to turn off sigma clipping within the background classes one would do, e.g.:

```
bkg = MedianBackground(sigma_clip=None)
```

PR #432, opened 2016-09-26
Soften API warning

eteq:
The current docs index page has a somewhat scary-sounding warning.  This PR updates it to be a bit softer, basically saying "we won't break the API unless there's a good reason."

Does this seem like reasonable wording (and something that's reasonable to keep), @larrybradley

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -23,7 +23,7 @@ Photutils
 Photutils at a Glance
 =====================
 
-**Photutils** is an in-development `affiliated package
+**Photutils** is an  `affiliated package
 <http://www.astropy.org/affiliated/index.html>`_ of `Astropy`_ to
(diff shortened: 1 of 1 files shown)


larrybradley:
:+1:  Yes, thanks.  That was a holdover from a few years ago, I think.  We don't want to scare anyone away!

larrybradley:
@eteq  Were you suggesting that we remove this note altogether?  I'm fine with that too.

eteq:
@larrybradley - nope I meant just as it stands in this PR.  I think it's probably good to make sure people know the API might change and use that as a way to try to engage users. I just don't think we want it as a scary "we might change this without warning" style statement.  I think from 0.3 on it will be more true, anyway.

larrybradley:
Thanks, @eteq!

PR #449, opened 2016-10-31
Minor language fixes in PSF docs

barentsen:
Minor language fixes in the new PSF docs attached.


larrybradley:
Thanks, @barentsen!

PR #451, opened 2016-11-04
Removes the psf_photometry function in favor of the PSFPhotometry classes

mirca:
This PR removes the psf_photometry functionality.

In addition, this PR also adds a `raise ValueError` when `flux_0` (initial guesses from flux) and `aperture_radius` are not given (and `aperture_radius` cannot be determined by `psf_model`). When this occurs, the photometry process is not able to determine initial guesses for fluxes.

Question: @bsipocz @eteq @hamogu should we maybe allow ``group_maker`` to be ``None``, for instance, in the case that the user just wants to perform photometry in a non crowded scenario?


eteq:
Looks like the conflict is from here: https://github.com/astropy/photutils/commit/3d64855a65bb09cd5001a2da3c82224f310bb4e0

But I think it'd be best to move the stuff for `DiscretePRF` to a `test_sandbox` or similar, though.  Other than that this looks good!

mirca:
@eteq Failures seem to be unrelated.

larrybradley:
Thanks, @mirca!

PR #452, opened 2016-11-04
Allow FittableImageModel to support oversampling

eteq:
This updates `FittableImageModel` (added in #413) to support oversampling.  It also beefs up the tests of that model to be a bit more complete.

cc @larrybradley


larrybradley:
The failing test looked like a transient failure.  I restarted it.

larrybradley:
Thanks, @eteq!

PR #456, opened 2016-11-07
Update 0.3 changelog

mirca:
I forgot to write the changelog entries for those PRs, which resulted in a misleading changelog for the 0.3 release. Sorry.

cc @eteq @larrybradley

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -62,13 +62,18 @@ New Features
 
   - Added ``filter_data`` function. [#398]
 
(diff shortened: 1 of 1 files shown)


larrybradley:
Thanks, @mirca!

eteq:
@larrybradley - I'm guessing we'll want to do a 3.1 at some point in the near future, so are we OK with that including this, even though it makes the changelogs appear inconsistent?

larrybradley:
Hopefully it won't be too confusing --  already merged it!  Sure we can do a 0.3.1 after adding the PSF docs.

PR #460, opened 2016-11-10
fix subtle oversampling bug

eteq:
This fixes a subtle but very important bug in how the `FittableImageModel` represents its center.  This is a pretty major functionality bug so it might be reason for a 0.3.1 release soonish...

cc @larrybradley


larrybradley:
👍 I'm merge as soon as CI completes.

PR #520, opened 2017-03-19
remove double colon before doctest-requires

migueldvb:
This prevents the `.. doctest-requires:: scipy, skimage` directive to appear as a code block in the HTML documentation.


larrybradley:
Thanks, @migueldvb!

PR #566, opened 2017-07-26
Catch warning about no detected sources in IterativelySubtractedPSFPhotometry

larrybradley:
`IterativelySubtractedPSFPhotometry` iterates until the star finders detect no more sources.  The star finders issue a warning when no sources are detected.  Thus, `IterativelySubtractedPSFPhotometry` will always raise this warning, even when everything worked perfectly fine.

This PR catches this warning after the first iteration of `IterativelySubtractedPSFPhotometry`.  In other words, if no stars are detected on the first iteration, the warning is still issued.  After that, the warning is suppressed.

Also, this warning is the reason why the Sphinx builds are failing for #562.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -91,6 +91,12 @@ API changes
   - The ``unit``, ``hdu``, ``wcs``, and ``wcsheader`` keywords in
     ``photutils.datasets`` functions were removed. [#527]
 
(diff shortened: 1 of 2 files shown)


larrybradley:
Thanks, @mirca!

PR #591, opened 2017-10-04
update terminology on discussion from PSF workshop

eteq:
This is an update to the "terminology" section of the docs, based on a variety of discussions from a PSF workshop held Oct 3/4, 2017.


docs/photutils/psf.rst
```diff
@@ -13,20 +13,33 @@ called "PSF photometry".
...
+Different astronomy sub-fields use the terms "PSF", "PRF", or related terms
+somewhat differently, especially when colloquial usage is taken into account.
+This package aims to be at the very least internally consistent, following the
+definitions described here.
+
+For this module we take Point Spread Function (PSF), or instrumental Point Spread
+Function (iPSF) to be the infinite resolution and infinite signal-to-noise flux
+distribution from a point source on the detector, after passing through optics,
+dust, atmosphere, etc. By contrast, the function describing the efficiency of
+the *pixels* of the detector is the Pixel Response Function (sometimes called
```
barentsen: Nitpicking: the words `describing the efficiency of the detector` makes it sound as if the PRF also folds in the flat-field, i.e. the intra- and inter-pixel sensitivity variations of the detector.  Is this the intended meaning?

karllark: I agree that the current terminology is a bit vague.  It would be good to explicitly state that the "Pixel Response Function" means the responsivity variations across a single pixel.

eteq: Good point - I was meaning just what @karllark says, but @barentsen is right it's unclear.  Will update in the next round.

PR #617, opened 2017-11-10
Fix deblending with sources not in the segmentation mask

### LINKED ISSUE #616: Deblending error
Trying the detection & deblending code I have the following error:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-33-11b338bc9d78> in <module>()
      1 from photutils import deblend_sources
      2 segm_deblend = deblend_sources(data, segm, npixels=5,
----> 3                                filter_kernel=kernel)

~/lib/photutils/photutils/segmentation/deblend.py in deblend_sources(data, segment_img, npixels, filter_kernel, labels, nlevels, contrast, mode, connectivity, relabel)
    135                              'in detect_sources and deblend_sources.  If '
    136                              'this issue persists, then please inform the '
--> 137                              'developers.'.format(label))
    138 
    139         if source_deblended.nlabels > 1:

ValueError: Deblending failed for source "83".  Please ensure you used the same pixel connectivity in detect_sources and deblend_sources.  If this issue persists, then please inform the developers.
```
I cannot share the image but will try to make a reproductible example or identify the issue. A first hint with the `source_deblended.data` and `source_segm.data` arrays below, it seems that the source in the bottom left corner appeared while deblending the big blob:
 
image 
image image

```
ipdb>  np.where(source_deblended.data.astype(bool) != source_segm.data.astype(bool))
(array([3, 4, 4, 4, 5, 5, 5, 6]), array([2, 1, 2, 3, 1, 2, 3, 2]))
```

saimn:
Fix #616.

To deblend a source a small stamp is created and can contain other
sources, depending on the shape of the source mask. These neighboring
sources are not masked so they are used in the deblending process. Given
that the data array is also used for flux computation before the
watershed, the most abvious solution is to put values outside the source
mask to zero. 

Maybe there is a better approach ? I will also try to make a test with a simulated image which reproduce the issue.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ New Features
 
   - Added a ``mask`` keyword to the ``detect_sources`` function. [#621]
 
+  - Fix ``deblend_sources`` when other sources are in the neighborhood. [#617]
+
 
 0.4 (2017-10-30)
 ----------------
--- a/photutils/segmentation/deblend.py
+++ b/photutils/segmentation/deblend.py
@@ -224,8 +224,11 @@ def _deblend_source(data, segment_img, npixels, nlevels=32, contrast=0.001,
         raise ValueError('Invalid connectivity={0}.  '
                          'Options are 4 or 8'.format(connectivity))
 
+    # Work on a copy of the data to avoid modifying the input image
+    data = data.copy()
     segm_mask = (segment_img.data > 0)
     source_values = data[segm_mask]
+    data[~segm_mask] = 0
     source_min = np.min(source_values)
     source_max = np.max(source_values)
     if source_min == source_max:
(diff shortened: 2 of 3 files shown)


saimn:
The result with the same test case as #616 (nice!):
image

saimn:
Added a test :
image

The test with negative values is failing, not sure what's the best solution here. Maybe just use the minimum values to fill the data instead of 0 ?

larrybradley:
Many thanks, @saimn!  Let me investigate that failing test.

larrybradley:
@saimn  I added a `mask` keyword to the `detect_sources` function (#621), which can be used to exclude these other sources in the data.  It's also a generally useful feature.

Once #621 is merged, the simple fix here is to add `mask=~segm_mask` to the call to `detect_sources` here: https://github.com/astropy/photutils/blob/master/photutils/segmentation/deblend.py#L256-L257

saimn:
@larrybradley - Sounds good, it's indeed a better solution. I will update the PR once #621 is merged, thanks.

saimn:
I rebased and updated the PR to use the new `mask` keyword from #621 .

larrybradley:
Thanks, @saimn!

PR #678, opened 2018-06-18
Fix typo in spline degree for Y axis

mcara:
`ky` argument (spline degree for the Y-axis) for `RectBivariateSpline` is incorrect: `ky` should be set to `degy` - no to `degx`.


larrybradley:
Thanks, @mcara!

PR #683, opened 2018-06-28
Used Polygon to implement  method for rectangular apertures

fdeugenio:
I have implemented the exact method for the rectangles, using the python [Polygon](https://pypi.org/project/Polygon3/) bindings to the [GPC Library](http://www.cs.man.ac.uk/~toby/alan/software/). Polygon now becomes a necessary dependence.
No new testing has been added, but I have enabled the existing tests inside the aperture module, because before they singled out RectangularAperture and RectangularAnnulus .

Again, some testing will fail, but I think this is due to upstream problems in photutils rather than my edits (please correct me if I am wrong).

To prove my hypothesis about the tests, I attach a screenshot of the test summary for a test done on the branch exactrect (left), as well as a test done on the master (right).
screenshot from 2018-06-28 13-20-42

### DIFF
--- a/photutils/aperture/circle.py
+++ b/photutils/aperture/circle.py
@@ -31,8 +31,7 @@ class CircularMaskMixin(object):
         ----------
         method : {'exact', 'center', 'subpixel'}, optional
             The method used to determine the overlap of the aperture on
-            the pixel grid.  Not all options are available for all
-            aperture types.  Note that the more precise methods are
+            the pixel grid. Note that the more precise methods are
             generally slower.  The following methods are available:
 
                 * ``'exact'`` (default):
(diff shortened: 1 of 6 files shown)


larrybradley:
@fdeugenio  Many thanks for this contribution.  Having an exact method for rectangular apertures is a nice addition!

However, let me explain why this change should not go into `photutils`.  The low-level Cython geometry functions (e.g `rectangular_overlap.pyx`) currently in `photutils` are moving upstream to the astropy `regions` package (https://github.com/astropy/regions).  The plan is that `regions` will eventually move to the `astropy` core package (soonish, hopefully).  The `geometry` functions have already been moved to `regions` and the very near term they will be deprecated in `photutils`.  Therefore, this change should go in `regions` instead of `photutils` (I'm sorry that there was no way for you to know this plan).  This is a nice contribution, so please submit this PR to the `regions` package.

fdeugenio:
Will try to re-submit to astropy

PR #698, opened 2018-08-02
Remove jupyter from RTD environment

larrybradley:
I don't think we need `jupyter` any longer in the RTD environment.  This should speed up the RTD builds some.

CC:  @bsipocz


larrybradley:
This provided no noticeable impact.  RTD is still failing due to timeouts. ☹️

PR #709, opened 2018-09-05
Return an empty Table instead of an empty list for consistency in find_peaks

larrybradley:
If the input data is an array of constant values, an empty `Table` is now returned instead of an empty list for consistency when no local peaks are found.

CC: @cdeil

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,14 @@
 0.6 (unreleased)
 ----------------
 
-- No changes yet
+API changes
+^^^^^^^^^^^
+
+- ``photutils.detection``
+
+  - The ``find_peaks`` function now returns an empty
+    ``astropy.table.Table`` instead of an empty list if the input data
+    is an array of constant values.  [#709]
 
 
 0.5 (2018-08-06)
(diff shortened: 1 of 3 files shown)


cdeil:
👍 , thanks!

FYI: I put a simple version in [gammapy.detect.find_peaks](http://docs.gammapy.org/dev/api/gammapy.detect.find_peaks.html) to avoid the extra dependency for just one function, and also to support a `gammapy.maps.Map` object as input directly.

The one thing I added was this
```
    # Remove non-finite values to avoid warnings or spurious detection
    data = image.data.copy()
    data[~np.isfinite(data)] = np.nanmin(data)
```
because the images we run the peak finder on often have borders of `NaN` values and our users don't like seeing warnings that they are "normal" and should be ignored; the comparison of data to threshold otherwise emits this from Numpy:
```
RuntimeWarning: invalid value encountered in greater
```

If you / the photutils users don't have `NaN` or complain about this, just leave as-is; I just mentioned it in case you might want to add that too.

larrybradley:
@cdeil  Yes, I think that's useful.  I will add it in a separate PR.

PR #717, opened 2018-09-18
Add quadratic centroid estimator

mirca:
This code implements a quadratic estimator for centroids in 2D arrays based on the work of [Vakili & Hogg (2016) ](https://arxiv.org/abs/1610.05873).

The fundamental idea is that most of the information to infer the centroids of a blob are located in the center pixels, and fitting a quadratic function would capture most of that information. In practice, [Vakili & Hogg (2016) ](https://arxiv.org/abs/1610.05873) showed that the information loss resulted by ignoring the values of other pixels is negligible in various scenarios of signal-to-noise ratio.

This method has been implemented for one dimensional data in https://github.com/richteague/bettermoments.

This implementation is a collaboration with @dfm, @nksaunders, @rodluger, @christinahedges, @afeinstein20, @megbedell, and @benmontet.

### DIFF
--- a/docs/centroids.rst
+++ b/docs/centroids.rst
@@ -17,6 +17,9 @@ centroid of a single source.  The centroid methods are:
 * :func:`~photutils.centroids.centroid_2dg`: Calculates the centroid
   by fitting a 2D Gaussian to the 2D distribution of the data.
 
+* :func:`~photutils.centroids.centroid_quadratic_2d`: Calculates the centroid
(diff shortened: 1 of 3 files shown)


larrybradley:
Thanks, @mirca!  I should be able to review this tomorrow.

mirca:
Thanks Larry, no rush!

mirca:
I'm going to close this as it has been added to https://github.com/KeplerGO/lightkurve

PR #722, opened 2018-09-24
Initial framework for SingleObjectModel within PSF Photometry

Onoddil:
Initial, backwards-compatible functionality of the SingleObjectModel within the PSF photometry framework. Specification documentation included, detailing the aim of the block, its input and output parameters, with a simple use case. Also included is a very brief description of SceneMaker, the future extension to SingleObjectModel, detailing its intent.

SingleObjectModel will allow for the potential relaxing of the assumption that all sources being fit with PSF photometry are point sources, with users able to provide custom classes which return the combined PSF and extended source profile model, allowing for the fitting of extended sources. Current implementation provides for a default, basic single point source profile, maintaining backwards compatibility with previous versions.


Onoddil:
Closing this PR to split it into several smaller pieces. See future PR for details.

PR #724, opened 2018-09-24
Initial framework for CullerAndEnder in Iterative PSF Photometry

Onoddil:
Initial framework added for handling the evaluation of the goodness-of-fit and premature ending of iterative PSF photometry fitting routines. Specification documentation added giving description of the class, input and output parameters and simple use case. Backwards compatible class set as default, in which no sources are evaluated for quality of fit with no sources being culled, and no consideration is given to whether iterative fitting should be ended early.


Onoddil:
Closing this PR to split it into several smaller pieces. See future PR for details.

PR #725, opened 2018-09-24
Fix to ePSF normalisation treatment

Onoddil:
Pull request to handle the fitting of effective PSFs in a manner consistent with that of the original Anderson & King (2000, PASP, 112, 1360) paper. Oversampled ePSF grid is no longer assumed to be a normalised image, but properly handles the normalisation in the original, undersampled grid. Currently the flux evaluation of ``EPSFModel`` is a simple ``np.sum(data)``, but the ePSF grid is an oversampling of data points where each point represents the fraction of light falling within a "normal" (i.e., undersampled) pixel centered at its x, y location relative to a point source. Thus the flux evaluation is currently overestimated by a factor of approximately oversampling^2. 

The main fix in this PR is then a change to ``EPSFModel``'s ``_compute_raw_image_norm`` function, in which only integer pixel values are picked to compute the normalization integral (or sum, in this case). Most of the subsequent changes are consistency fixes for whether pixel indices should be referenced by their undersampled (i.e., 0 to 25 for a 25x25 cutout of a star from an observation) or oversampled (i.e., 0 to 101 for a 4x oversampled ePSF grid) values. For example, ``EPSFBuilder`` now has  ``_resample_residual`` first use the oversampled ``x`` and ``y`` indices to compute the (e.g.) 0.25x0.25 pixel bin the sample falls into, and then uses the undersampled ``x`` and ``y`` values for evaluating the ePSF at the finer sample offsets.

``EPSFModel`` is no longer a subclass of ``FittableImageModel`` but itself an additional subclass of ``Fittable2DModel``, as ``EPSFModel`` overloads a significant number of ``FittableImageModel`` functions so it felt less confusing to simply define it as its own additional subclass. 

This PR also removes the factor of ``oversampling`` required in the star fitting routines -- such as ``register_epsf`` in ``EPSFStar`` -- to correct for their fluxes, which allows for the two routines to be used separately from one another. This then gives consistent results between the fitting done in ``EPSFFitter`` and the fluxes derived from using the ePSF model in, for example, ``BasicPSFPhotometry``.

Also included is the treatment of grid centroiding as ``centroid_epsf``, as detailed by Anderson & King (2000) using the asymmetry of the grid at maximum gradient. Finally, ``pixel_scale`` is removed as a parameter accepted by ``EPSFModel``, with ``oversampling`` defined by its input rather than allowing for the inclusion of two pixel scales and calculating the ratio between them, streamlining the setup process and reducing confusion from a user point of view.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,23 @@ Other Changes and Additions
   keywords. [#726]
 
 
+Bug Fixes
+^^^^^^^^^
+
+- ``photutils.psf``
+
+  - Fixed normalization of ePSF model, now correctly normalizing on
+    undersampled pixel grid.
+
+Other Changes and Additions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Updated background and detection functions that call
+  ``astropy.stats.SigmaClip`` or ``astropy.stats.sigma_clipped_stats``
+  to support both their ``iters`` (for astropy < 3.1) and ``maxiters``
+  keywords. [#726]
+
+
 0.5 (2018-08-06)
 ----------------
 
--- a/photutils/centroids/core.py
+++ b/photutils/centroids/core.py
@@ -557,27 +557,26 @@ def centroid_epsf(data, mask=None, oversampling=4, shift_val=0.5):
     psi_pos_x = data[y_0, x_0 + x_shiftidx]
     psi_pos_x_m1 = data[y_0, x_0 + x_shiftidx - 1]
     psi_pos_x_p1 = data[y_0, x_0 + x_shiftidx + 1]
-    dpsi_pos_x = (psi_pos_x_p1 - psi_pos_x_m1) / 2.
+    dpsi_pos_x = np.abs(psi_pos_x_p1 - psi_pos_x_m1) / 2.
     # psi_E(-0.5, 0.0) and derivative components.
     psi_neg_x = data[y_0, x_0 - x_shiftidx]
     psi_neg_x_m1 = data[y_0, x_0 - x_shiftidx - 1]
     psi_neg_x_p1 = data[y_0, x_0 - x_shiftidx + 1]
-    dpsi_neg_x = (psi_neg_x_p1 - psi_neg_x_m1) / 2.
+    dpsi_neg_x = np.abs(psi_neg_x_p1 - psi_neg_x_m1) / 2.
 
-    x_shift = (psi_pos_x - psi_neg_x) / (dpsi_pos_x - dpsi_neg_x)
+    x_shift = (psi_pos_x - psi_neg_x) / (dpsi_pos_x + dpsi_neg_x)
 
     # psi_E(0.0, 0.5) and derivatives.
     psi_pos_y = data[y_0 + y_shiftidx, x_0]
     psi_pos_y_m1 = data[y_0 + y_shiftidx - 1, x_0]
     psi_pos_y_p1 = data[y_0 + y_shiftidx + 1, x_0]
-    dpsi_pos_y = (psi_pos_y_p1 - psi_pos_y_m1) / 2.
+    dpsi_pos_y = np.abs(psi_pos_y_p1 - psi_pos_y_m1) / 2.
     # psi_E(0.0, -0.5) and derivative components.
     psi_neg_y = data[y_0 - y_shiftidx, x_0]
     psi_neg_y_m1 = data[y_0 - y_shiftidx - 1, x_0]
     psi_neg_y_p1 = data[y_0 - y_shiftidx + 1, x_0]
-    dpsi_neg_y = (psi_neg_y_p1 - psi_neg_y_m1) / 2.
+    dpsi_neg_y = np.abs(psi_neg_y_p1 - psi_neg_y_m1) / 2.
 
-    y_shift = (psi_pos_y - psi_neg_y) / (dpsi_pos_y - dpsi_neg_y)
-
+    y_shift = (psi_pos_y - psi_neg_y) / (dpsi_pos_y + dpsi_neg_y)
 
     return x_0+x_shift, y_0+y_shift
--- a/photutils/centroids/core.py
+++ b/photutils/centroids/core.py
@@ -15,11 +15,6 @@ from astropy.modeling.models import (Gaussian1D, Gaussian2D, Const1D,
 from astropy.modeling.fitting import LevMarLSQFitter
 from astropy.nddata.utils import overlap_slices
 from astropy.utils.exceptions import AstropyUserWarning
-try:
-    from ..psf import _interpolate_missing_data
-except ImportError:
-    pass
-
 
 __all__ = ['GaussianConst2D', 'centroid_com', 'gaussian1d_moments',
            'fit_2dgaussian', 'centroid_1dg', 'centroid_2dg',
@@ -533,13 +528,6 @@ def centroid_epsf(data, mask=None, oversampling=4, shift_val=0.5):
     if shift_val <= 0:
         raise ValueError('shift_val must be a positive number.')
 
-    badidx = ~np.isfinite(data)
-    if np.any(badidx):
-        warnings.warn('Input data contains input values (e.g. NaNs or infs), '
-                      'which were automatically set to the '
-                      'surrounding pixels.', AstropyUserWarning)
-        data = _interpolate_missing_data(data, badidx)
-
     if oversampling is None:
         raise ValueError('oversampling must be supplied.')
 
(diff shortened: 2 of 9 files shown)


eteq:
@Onoddil - a couple big things before a full review:

1. Why is `centroid_epsf` in this PR? it doesn't seem to be related to the epsf normalization? So maybe it should be moved to a separate PR? This one is already quite large which is why it's been taking a while for @larrybradley or I do get to a proper review.

2. The change to `Fittable2DModel` (and various other changes here) are a major API break.  This code has already been released in a version of photutils and we know it has several users already, so we want to be a it cautious of just changing things under people's feet without cause. More generally, I don't understand why the `Fittable2DModel` change was made at all. @larrybradley might correct me on this, but I think `FittableImageModel` has several methods that *are* still used by the EPSF machinery.  In particular I note that `pixel_scale` is missing, which is the whole point of `EPSFModel` in the first place...?

Onoddil:
@eteq 
1) I was unsure of whether to include ``centroid_epsf``. The question of its relevance is whether the previous methods of centroiding (``centroid_com`` using centre of mass, for example) are applicable to the new, non-normalised image grid. I did some rough tests and got more or less the same answer with both methods, so perhaps it should be stripped out into a separate request, but I had included it as it is an addition to the ePSF fitting methodology. If this should be "normalisation only" PR and "a new way to centroid" PR then that's fine; I did feel they were linked sufficiently to be included in the same pull request but if that makes it too unwieldy so be it.

2) It is possible to revert ``EPSFModel`` to being a subclass of ``FittableImageModel``, but a large fraction of functions are overloaded (both the normalisation calculation routines and ``evaluate`` change, for example), making the base/sub class distinction a bit blurry. In addition, there are several functions that are no longer used, which would then be callable from the base ``FittableImageModel`` class but not serve any purpose for the new formalism. 
``pixel_scale`` being removed was something I discussed with @larrybradley briefly, apologies for forgetting to add its removal to the PR description. ``pixel_scale`` was implemented as a way to compute ``oversampling`` indirectly (i.e., you have regular pixel scale ``x`` and ePSF pixel scale ``y`` so ``oversampling`` is ``x/y``); however, this is unnecessary as the point of ``oversampling`` is simply to define the extent to which the undersampled grid is probed, with an oversampling of 4 likely appropriate for almost all purposes. It is therefore confusing and unnecessary to have two ways to define the number of ePSF grid points evaluated between each "regular", undersampled image pixel centers. I think in this instance an API change is merited to remove the conflicting nature of the two input arguments and streamline function setup.
I do not know where ``FittableImageModel`` is used in the rest of the codebase, and (as I said above) ``EPSFModel`` can be (re)made a subclass of it, but I am unsure of the purpose of ``FittableImageModel`` -- ``IntegratedGaussianPRF`` is a ``Fittable2DModel``, for instance. If backwards-compatibility is more important here then I will revert ``EPSFModel`` back to being a subclass with vestigial class functions.

Onoddil:
With regards to the requirement that ``EPSFModel`` be reverted back to a ``FittableImageModel`` subclass, @eteq , I just wanted to further expand on the decision to split the subclass into its own class and see whether the changes that would be necessary to bring the two classes back into agreement are better or worse than an API change.

The main changes to the ``EPSFModel``, as I mentioned, are those surrounding the handling of the normalisation of the empirical ePSF pixel values. Thus ``EPSFModel`` overloads ``_compute_raw_image_norm`` and ``_compute_normalization``, and removes the secondary confusion with ``self._img_norm`` and ``self._normalization_constant``, simply re-normalising the data array if ``Normalize=True``. 

However, the issue is that ``EPSFModel`` requires an additional parameter, ``self._norm_radius``, defining a circular number of pixels inside which to calculate the normalisation on undersampled, 'regular' pixels (i.e., every Nth ePSF pixel for oversampling=N). Thus when ``EPSFModel`` calls ``super().__init()`` and gets to https://github.com/Onoddil/photutils/blob/d77d1efba91c86d91c8576d9aa897b5dc9b8d5ed/photutils/psf/models.py#L146-L151 it will never be able to pass ``self._norm_radius``, and thus in the new ``EPSFModel`` that part of ``__init__`` is (ignoring changes to catchalls which could potentially be propagated backwards into ``FittableImageModel``) https://github.com/Onoddil/photutils/blob/d77d1efba91c86d91c8576d9aa897b5dc9b8d5ed/photutils/psf/models.py#L551-L562 

I believe all other functions could be absorbed back into ``FittableImageModel`` and then overloaded if backwards compatibility with the API is paramount, but the initialisation of ``self._img_norm``, and ``flux`` are handled sufficiently differently in the two classes that those code snippets would have to be pulled out into a function that ``EPSFModel`` can then overload. My question is then: is this change better (allowing for ``EPSFModel`` to become a ``FittableImageModel`` once again) or worse (making changes to both ``EPSFModel`` AND ``FittableImageModel``)? I chose option 1, thinking that changing two classes was worse than changing just the one being affected, but if option 2 is the preferred choice then that is relatively straightforward to change. Please keep this in mind when reviewing the code (cc @larrybradley if you happen to be the reviewer).

Onoddil:
It is also likely advisable that PR #733 be looked at and reviewed before this PR, as there is a small amount of overlap in some code to ensure that CI builds continued to run. If/when the smaller PR is accepted I can rebase this branch to remove the code unnecessary for this change specifically (as well as fix the current minor conflict), which should hopefully make it a bit easier to review.

Onoddil:
As per a discussion had with @eteq I have reversed the splitting of ``EPSFModel`` from a subclass of ``FittableImageModel``. To do so a minor change to ``FittableImageModel.__init__`` was necessary, placing the initialisation of the normalisation and setting of ``flux`` and ``self._img_norm`` into a small internal functionn to allow for ``EPSFModel`` to overload it and pass ``self.norm_radius`` as an additional parameter. This is a cosmetic change to ``FittableImageModel``, however, simply moving a few lines of code from the ``__init__`` call to a small function which returns ``flux`` as the only changed variable within the function scope.

Given that this is a minor change, setting subclass dependency and removing duplicate code, the results should be entirely consistent; one minor change to ``epsf.py`` was required to fix a small bug in the setting of ``origin``, revealed when ``EPSFModel.origin``'s ``setter`` function changed slightly; however, this change is also relatively cosmetic, as the same calculation was being performed in both places, modulo the now-revealed numerical calculation mistake.

This does result, as mentioned above, in a few 'dangling' functions -- given the re-defining of the normalisation, ``self.normalized_data`` is overloaded to simply return ``data``, to avoid the user accidentally calling it and getting conflicting answers, for instance -- but does overall reduce the codebase by eliminating truly duplicated code. I therefore believe that the most recent commit (b15b7e9) resolves the ``FittableImageModel`` subclass issue discussed above.

Additionally, I just re-mention that #733 should be checked and (after any changes) accepted before this PR as this PR uses some code from #733 -- once the first branch is merged I will rebase, resolving the conflict in ``test_epsf`` at the same time.

larrybradley:
@Onoddil  Thanks for the contribution and sorry I haven't given this a thorough review yet.  I see there are several changes in this PR beyond the ePSF normalization.  Can you please make a separate PR for each significant change?  The 1. centroiding, 2. removal of pixel scale, 3. Gaussian PRF tests, and 4. ePSF normalization all should be separate PRs.  That will help me give this PR a proper review.  If there is time, some of these changes could go into the upcoming 0.6 release (which I need to get out ASAP for JWST; within a week).

hamogu:
I just run into the issue of normalization ass user. I was looking to open an issue, and was happy to find that there's even a PR to fix it. For now, I'll just manually correct all fitted fluxes by the oversampling factor. I suspected that there might be a bug like that, but I was reluctant to guess that factor from just looking at my images. So, thanks for your detailed discussion here which makes me feel good about doing that manual correction. 

However, it would obviously be even better to get this PR merged (either in this form or split into smaller components).

eteq:
Thanks for the ping @hamogu - hopefully there's a path forward below!

@larrybradley - @Onoddil and I discussed this out-of-band and it looks like:
1. your 3 above should just go away since it's pretty much already in
2. @Onoddil will do the splitting as you asked, but in 3 separate follow-on PRs since rebasing this one will be a bear
3. There *is* an intrinsic connection between 1 and 4 as it turns out: the normalization requires accounting for the pixel oversampling, which the new `centroid_epsf` does. What I'm thinking @Onoddil should do is add the oversampling bit to the `centroid_com` and use that as part of 4, since that's the minimal change required.  Then 1 will build off that and switch the default over to `centroid_epsf`.

Onoddil:
Closing this PR now, and will re-open as three subsequent PRs as discussed above. This thread is now basically just for reference and context; please see new PRs for an updated codebase.

PR #733, opened 2018-10-03
Updated test_photometry to correctly use IntegratedGaussianPRF

Onoddil:
Changed test image creation in ``test_photometry`` to correctly create discrete pixel images, instead of evaluating the Gaussian PDF at the pixel centre. The PR adds a new function, ``make_gaussian_prf_sources_image`` which is analogous to the original ``make_gaussian_sources_image`` except it calls ``IntegratedGaussianPRF`` instead of ``Gaussian2D`` as the input model. Minor changes were made to the various tests as ``IntegratedGaussianPRF`` takes differing inputs -- ``x_mean`` becomes ``x_0``, etc.

This fixes the issue seen where fit and input fluxes were incompatible, as two differing models were being fit and producing differing results. Thus the accuracy of several ``assert_allclose`` statements could be improved with the conflict resolved.

### DIFF
--- a/photutils/datasets/tests/test_make.py
+++ b/photutils/datasets/tests/test_make.py
@@ -20,20 +20,20 @@ try:
 except ImportError:
     HAS_SCIPY = False
 
-TABLE = Table()
-TABLE['flux'] = [1, 2, 3]
-TABLE['x_mean'] = [30, 50, 70.5]
-TABLE['y_mean'] = [50, 50, 50.5]
-TABLE['x_stddev'] = [1, 2, 3.5]
-TABLE['y_stddev'] = [2, 1, 3.5]
-TABLE['theta'] = np.array([0., 30, 50]) * np.pi / 180.
+SOURCE_TABLE = Table()
+SOURCE_TABLE['flux'] = [1, 2, 3]
+SOURCE_TABLE['x_mean'] = [30, 50, 70.5]
+SOURCE_TABLE['y_mean'] = [50, 50, 50.5]
+SOURCE_TABLE['x_stddev'] = [1, 2, 3.5]
+SOURCE_TABLE['y_stddev'] = [2, 1, 3.5]
+SOURCE_TABLE['theta'] = np.array([0., 30, 50]) * np.pi / 180.
 
-TABLE2 = Table()
-TABLE2['x_0'] = [30, 50, 70.5]
-TABLE2['y_0'] = [50, 50, 50.5]
+SOURCE_TABLE_PRF = Table()
+SOURCE_TABLE_PRF['x_0'] = [30, 50, 70.5]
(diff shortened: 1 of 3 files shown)


photutils/datasets/tests/test_make.py
```diff
@@ -22,6 +27,13 @@
 TABLE['y_stddev'] = [2, 1, 3.5]
 TABLE['theta'] = np.array([0., 30, 50]) * np.pi / 180.
 
+TABLE2 = Table()
```
eteq: What about `TABLE_PRF` instead of `TABLE2`?  While you're at it I suppose you could change them both to something more like `SOURCES_TABLE` since "table" is pretty ambiguous.

photutils/psf/tests/test_photometry.py
```diff
@@ -580,7 +576,7 @@ def test_psf_photometry_moffat():
     for n in ['x', 'y']:
         assert_allclose(f[n + '_0'], f[n + '_fit'], rtol=1e-3)
     # image was created with a gaussian, so flux won't match exactly
-    assert_allclose(f['flux_0'], f['flux_fit'], rtol=1)
+    assert_allclose(f['flux_0'], f['flux_fit'], rtol=1e-1)
```
eteq: Why this change?  shouldn't this be unaffected by the rest of the changes in this PR?

photutils/datasets/tests/test_make.py
Onoddil: I have fixed the table names to be better. I liked ``SOURCES_TABLE`` so we now have ``SOURCES_TABLE`` and ``SOURCES_TABLE_PRF``.

photutils/psf/tests/test_photometry.py
Onoddil: This is the fix I (perhaps too briefly) mention above, with the improvements to ``assert_allclose``. Previously the incompatibility between the test image creation routine and the fitting routine meant that the test failed, essentially. This meant that, in order to pass the test, the tolerance had to be inflated significantly -- in this case with a tolerance of 100%, or do the input and output flux agree to a factor between 0.5 and 2. As the two functions (PDF Gaussian evaluated at pixel center vs integration over a pixel) weren't _that_ wrong this criterion was passed, but more or less resulted in a test that was wrong, but meaningless; a test with ``rtol=1`` might as well just not have been run in the first place.

I lowered this to 10% so that the test can now actually fail if incompatible changes are made to either routine which affect their results by a non-negligible but not massive amount. Unfortunately I now see that I did not remove the explaining comment for such a high tolerance, which should probably be removed to avoid confusion in the future.

eteq:
(note I think this was ready but needed a rebase, so I just pushed that up now and I will merge if the tests pass.)

PR #739, opened 2018-10-08
Added check for NDData.uncertainty being StdDevUncertainty array in construction of EPSFStar object

### LINKED ISSUE #734: ePSF weights are passed as data.uncertainty
When creating ``EPSFStar`` instance in ``extract_stars`` weights can be pulled from ``data.uncertainty.array`` but are never converted to weights, with ``weights`` simply being the quoted uncertainty. This is then propagated into ``EPSFFitter._fit_star`` where weights can be passed to ``fitter``, which would lead to the case that lower quality data would receive a higher relative weighting than high signal-to-noise data.

Onoddil:
Added the option for the weights in ``EPSFStar`` objects to be passed from ``StdDevUncertainty`` NDData arrays, extending from the original ``weights`` ``uncertainty_type`` option. Also added brief comment describing the origin of ``weights`` in EPSFFitter ``fitter``. Closes #734

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -112,6 +112,9 @@ Other Changes and Additions
   to support both their ``iters`` (for astropy < 3.1) and ``maxiters``
   keywords. [#726]
 
+- Added the option for ``EPSFStar`` to have ``weights`` based on an
+  ``astropy.nddata.StdDevUncertainty`` object.
+
 
 0.5 (2018-08-06)
 ----------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -113,7 +113,7 @@ Other Changes and Additions
   keywords. [#726]
 
 - Added the option for ``EPSFStar`` to have ``weights`` based on an
-  ``astropy.nddata.StdDevUncertainty`` object.
+  ``astropy.nddata.StdDevUncertainty`` object. [#739]
 
 
 0.5 (2018-08-06)
--- a/photutils/psf/epsf_stars.py
+++ b/photutils/psf/epsf_stars.py
@@ -784,8 +784,20 @@ def _extract_stars(data, catalog, size=(11, 11), use_xy=True):
     if data.uncertainty is None:
         weights = np.ones_like(data.data)
     else:
+        # If NDData instance has been set to a custom uncertainty type with
+        # type ``weights`` then take the data as they are
         if data.uncertainty.uncertainty_type == 'weights':
             weights = np.asanyarray(data.uncertainty.array, dtype=np.float)
+        # Else, if astropy.nddata.StdDevUncertainty has been implemented for
+        # the data uncertainties we can take the inverse standard deviation as
+        # the weight. Before assigning the weights we must ignore zero
+        # variance data, however, setting the weights for these data points to
+        # zero.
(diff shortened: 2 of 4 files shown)


photutils/psf/epsf_stars.py
```diff
@@ -772,8 +772,20 @@ def _extract_stars(data, catalog, size=(11, 11), use_xy=True):
...
+        # If NDData instance has been set to a custom uncertainty type with
+        # type ``weights`` then take the data as they are
         if data.uncertainty.uncertainty_type == 'weights':
             weights = np.asanyarray(data.uncertainty.array, dtype=np.float)
+        # Else, if astropy.nddata.StdDevUncertainty has been implemented for
+        # the data uncertainties we can take the inverse standard deviation as
+        # the weight. Before assigning the weights we must ignore zero
+        # variance data, however, setting the weights for these data points to
+        # zero by forcing their standard deviations to a large value.
+        elif data.uncertainty.uncertainty_type == 'std':
```
eteq: Note that in the upcoming astropy release (3.1) it looks to me like there will be an easier way to "normalize" other uncertainties into std.  We'll want to do that for this option I'd think, but no point in doing that until the astropy release is available for testing.

CHANGES.rst
```diff
@@ -36,6 +36,9 @@ Other Changes and Additions
   to support both their ``iters`` (for astropy < 3.1) and ``maxiters``
   keywords. [#726]
 
+- Added the option for ``EPSFStar`` to have ``weights`` based on an
+  ``astropy.nddata.StdDevUncertainty`` object.
```
eteq:
```suggestion
  ``astropy.nddata.StdDevUncertainty`` object. [#739]
```

changelog entries should have the relevant PR included at the end

photutils/psf/epsf_stars.py
Onoddil: So is this a "accept this and we'll revisit it later" or a "hold off on this until after the release is available for testing"? I think it would be good in the future to expand this quite a fair amount to accept various kinds of uncertainty (perhaps parsing them through some convenience function to produce an internally accepted 'standard' type), and it sounds like this will help with that down the line.

Onoddil:
Have made the changes requested @eteq - I note however that astropy 3.1 has been released subsequently and therefore the new changes to uncertainties must be updated; I think it makes sense to merge this PR as a base to build the new functionality for std normalisation onto, but if that should be built into this PR before acceptance then let me know.

photutils/psf/epsf_stars.py
astrofrog: Astropy 3.1 is out, so I've assigned @mwcraig to this issue in case he can help with this :)

mwcraig: Right now what you could is something like this:

```python
uncertainty_conversion_to_std = {
     'std': lambda x: x,  # do nothing
     'var': np.sqrt,  # square root of the variance is the standard dev
     'ivar': lambda x: 1 / np.sqrt(x)   # invert and square root to get standard dev
}

uncert = data.uncertainty.array.copy()
std_dev = uncertainty_conversion_to_std[data.uncertainty.type](uncert)
```

(note: I haven't tested this code but hopefully it conveys the idea).

PR #740, opened 2018-10-08
Add fallback aperture_radius to psf photometry

### LINKED ISSUE #728: About PSF model
Hi,
When I got the PSF model from EPSFBuilder. Could the model be used as the parameter in IterativelySubtractedPSFPhotometry: psf_model? It raise the error lile:
File "/home/yf/.conda/envs/python3/lib/python3.6/site-packages/photutils/aperture/circle.py", line 124, in __init__
    if r < 0:
TypeError: '<' not supported between instances of 'NoneType' and 'int'
Is that mean the EPSFBuilder does not return a proper 2D PSF model, or I made mistakes somewhere.

Onoddil:
Added a fallback ``aperture_radius`` to BasicPSFPhotometry and IterativelySubtractedPSFPhotometry for non-Gaussian PSF models that do not have either a sigma or FWHM attribute. If neither are found, routines fall back to the smaller of the two ``fitshape`` lengths, using the available (e)PSF model pixels to compute an aperture radius flux. Closes #728

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -84,6 +84,9 @@ Bug Fixes
   - Fixed a bug that caused photometry to fail on an ``EPSFmodel`` with
     multiple stars in a group. [#1135]
 
+  - Added a fallback ``aperture_radius`` for PSF models without a FWHM
+    or sigma attribute, raising a warning. [#740]
+
 - ``photutils.segmentation``
 
   - Fixed ``SourceProperties`` ``local_background`` to work with
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -256,6 +256,21 @@ class BasicPSFPhotometry:
             elif hasattr(self.psf_model, 'sigma'):
                 self.aperture_radius = (self.psf_model.sigma.value *
                                         gaussian_sigma_to_fwhm)
+            # If PSF model doesn't have FWHM or sigma value -- as it
+            # is not a Gaussian; most likely because it's an ePSF --
+            # then we fall back on fitting a circle of the average
+            # size of the fitting box. As ``fitshape`` is the width
+            # of the box, we need (width-1)/2 as the radius.
+            else:
+                self.aperture_radius = float(np.amin((np.asanyarray(
+                                             self.fitshape) - 1) / 2))
+                warnings.warn('aperture_radius is None and could not '
+                              'be determined by psf_model. Setting '
+                              'radius to the smallest fitshape size. '
+                              'If fitshape is significantly larger than '
+                              'the psf_model core lengthscale, consider '
+                              'supplying a specific aperture_radius.',
+                              AstropyUserWarning)
 
         if self.aperture_radius is None:
             if init_guesses is None:
@@ -741,6 +756,21 @@ class IterativelySubtractedPSFPhotometry(BasicPSFPhotometry):
                 elif hasattr(self.psf_model, 'sigma'):
                     self.aperture_radius = (self.psf_model.sigma.value *
                                             gaussian_sigma_to_fwhm)
+                # If PSF model doesn't have FWHM or sigma value -- as it
+                # is not a Gaussian; most likely because it's an ePSF --
+                # then we fall back on fitting a circle of the average
+                # size of the fitting box. As ``fitshape`` is the width
+                # of the box, we need (width-1)/2 as the radius.
+                else:
+                    self.aperture_radius = float(np.amin((np.asanyarray(
+                                                 self.fitshape) - 1) / 2))
+                    warnings.warn('aperture_radius is None and could not '
+                                  'be determined by psf_model. Setting '
+                                  'radius to the smallest fitshape size. '
+                                  'If fitshape is significantly larger than '
+                                  'the psf_model core lengthscale, consider '
+                                  'supplying a specific aperture_radius.',
+                                  AstropyUserWarning)
 
             output_table = self._do_photometry()
         return output_table
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -258,14 +258,18 @@ class BasicPSFPhotometry:
                                         gaussian_sigma_to_fwhm)
             # If PSF model doesn't have FWHM or sigma value -- as it
             # is not a Gaussian; most likely because it's an ePSF --
-            # then we fall back on fitting a circle of the average
-            # size of the fitting box. As ``fitshape`` is the width
-            # of the box, we need (width-1)/2 as the radius.
-            else:
+            # and if fluxes were not supplied in ``init_guesses`` then we fall
+            # back on fitting a circle of the average size of the fitting box.
+            # As ``fitshape`` is the width of the box, we need (width-1)/2 as
+            # the radius.
+            elif (init_guesses is not None and
+                  'flux_0' not in init_guesses.colnames):
                 self.aperture_radius = float(np.amin((np.asanyarray(
                                              self.fitshape) - 1) / 2))
-                warnings.warn('aperture_radius is None and could not '
-                              'be determined by psf_model. Setting '
+                warnings.warn('init_guesses were input, but the flux_0 column '
+                              'was not provided. Initial fluxes cannot be '
+                              'calculated as aperture_radius is None and '
+                              'could not be determined by psf_model. Setting '
                               'radius to the smallest fitshape size. '
                               'If fitshape is significantly larger than '
                               'the psf_model core lengthscale, consider '
@@ -736,7 +740,10 @@ class IterativelySubtractedPSFPhotometry(BasicPSFPhotometry):
                 # is not a Gaussian; most likely because it's an ePSF --
                 # then we fall back on fitting a circle of the average
                 # size of the fitting box. As ``fitshape`` is the width
-                # of the box, we need (width-1)/2 as the radius.
+                # of the box, we need (width-1)/2 as the radius. As opposed to
+                # ``BasicPSFPhotometry``, we have to fallback on this even if
+                # fluxes were given in ``init_guesses``, as this will be used
+                # in subsequent iterations.
                 else:
                     self.aperture_radius = float(np.amin((np.asanyarray(
                                                  self.fitshape) - 1) / 2))
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -272,30 +272,6 @@ class BasicPSFPhotometry:
                               'supplying a specific aperture_radius.',
                               AstropyUserWarning)
 
-        if self.aperture_radius is None:
-            if init_guesses is None:
-                raise ValueError('aperture_radius was not input and could '
-                                 'not be determined by the input psf_model '
-                                 '(e.g., an EPSFModel).  For tabular PSF '
-                                 'models, you must input the aperture_radius '
-                                 'keyword.  For analytical PSF models, you '
-                                 'must either input the aperture_radius '
-                                 'keyword or define a fwhm or sigma '
-                                 'attribute on your input psf_model.')
-
-            if (init_guesses is not None and
-                    'flux_0' not in init_guesses.colnames):
-                raise ValueError('init_guesses were input, but the "flux_0" '
-                                 'column was not present.  Initial fluxes '
-                                 'cannot be calculated because '
-                                 'aperture_radius must was not input and '
-                                 'could not be determined by the input '
-                                 'psf_model (e.g., an EPSFModel).  For '
-                                 'analytical PSF models, you must either '
-                                 'input the aperture_radius keyword or '
-                                 'define a fwhm or sigma attribute on your '
-                                 'input psf_model.')
-
         if init_guesses is not None:
             # make sure the code does not modify user's input
             init_guesses = init_guesses.copy()
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -192,6 +192,35 @@ class BasicPSFPhotometry:
 
         return self._residual_image
 
+    def set_aperture_radius(self):
+        """
+        Sets the fallback aperture radius for initial flux calculations
+        in cases where no flux is supplied for a given star.
+        """
+
+        if hasattr(self.psf_model, 'fwhm'):
+            self.aperture_radius = self.psf_model.fwhm.value
+        elif hasattr(self.psf_model, 'sigma'):
+            self.aperture_radius = (self.psf_model.sigma.value *
+                                    gaussian_sigma_to_fwhm)
+        # If PSF model doesn't have FWHM or sigma value -- as it
+        # is not a Gaussian; most likely because it's an ePSF --
+        # then we fall back on fitting a circle of the average
+        # size of the fitting box. As ``fitshape`` is the width
+        # of the box, we need (width-1)/2 as the radius.
+        else:
+            self.aperture_radius = float(np.amin((np.asanyarray(
+                                         self.fitshape) - 1) / 2))
+            warnings.warn('aperture_radius is None and could not '
+                          'be determined by psf_model. Setting '
+                          'radius to the smallest fitshape size. '
+                          'This aperture radius will be used if '
+                          'initial fluxes require computing for any '
+                          'input stars. If fitshape is significantly '
+                          'larger than the psf_model core lengthscale, '
+                          'consider supplying a specific aperture_radius.',
+                          AstropyUserWarning)
+
     def __call__(self, image, init_guesses=None):
         """
         Performs PSF photometry. See `do_photometry` for more details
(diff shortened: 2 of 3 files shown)


Onoddil:
This PR also fixes the issue where ``IterativelySubtractedPSFPhotometry`` passed ``bkg_estimator=None`` does not automatically set ``self._residual_image`` at an initial pass, and thus finder callables will fail with ``self._residual_image=None`` as the image to find sources in.

photutils/psf/tests/test_photometry.py
```diff
@@ -498,10 +499,68 @@ def test_aperture_radius_value_error():
                                     fitshape=7)
 
     intab = Table(data=[[1], [1]], names=['x_0', 'y_0'])
-    with pytest.raises(ValueError) as err:
+    with pytest.warns(AstropyUserWarning):
         basic_phot(image=image, init_guesses=intab)
 
-    assert 'aperture_radius is None' in str(err.value)
+
+def tophatfinder(image):
```
eteq: minor and debatable (i.e., if you disagree I won't push it), but I would say it makes more sense to put this inside the test function below instead of at the module-level, since it's specific to that one test.

photutils/psf/tests/test_photometry.py
```diff
@@ -498,10 +499,68 @@ def test_aperture_radius_value_error():
...
+    for x, y, f in zip(x0, y0, f0):
+        img[x+2:x+5, y+2:y+5] = f/9
+
+    basic_phot = BasicPSFPhotometry(group_maker=DAOGroup(2),
+                                    bkg_estimator=None, psf_model=prf,
+                                    fitshape=7)
+
+    intab = Table(data=[[19.6, 34.9, 37], [4.5, 40.1, 19.6]],
+                  names=['x_0', 'y_0'])
+    with pytest.warns(AstropyUserWarning):
```
eteq: Here (and below), I think you should use `warns` with the `match` keyword (See https://docs.pytest.org/en/latest/warnings.html#warns) because you can never be sure that there isn't some other random astropyuserwarning you're accidentally catching if you don't check that the message is the right onw.

photutils/psf/photometry.py
```diff
@@ -651,13 +656,27 @@ def do_photometry(self, image, init_guesses=None):
...
                 elif hasattr(self.psf_model, 'sigma'):
                     self.aperture_radius = (self.psf_model.sigma.value *
                                             gaussian_sigma_to_fwhm)
+                # If PSF model doesn't have FWHM or sigma value -- as it
+                # is not a Gaussian; most likely because it's an ePSF --
+                # then we fall back on fitting a circle of the average
+                # size of the fitting box. As ``fitshape`` is the width
+                # of the box, we need (width-1)/2 as the radius.
+                else:
+                    self.aperture_radius = float(np.amin((np.asanyarray(
```
eteq: I'm not certain about this, but I thought there was an option for `fitshape` to be None (meaning "fit the whole image")?  If so, you should catch that case and explicitly raise it as an exception (probably a `ValueError`?) that tells the user that it's impossible to determine the aperture

Onoddil: ``fitshape`` has as its setter
https://github.com/Onoddil/photutils/blob/54cbb50a834e14320b81c0414daf0339a3414cf2/photutils/psf/photometry.py#L132-L152
with its input requirements being ``fitshape : int or length-2 array-like``, so I do not believe that there is a ``fitshape=None`` option currently.

However, that _is_ a good idea, and one that may merit implementation, at which point this exception would be needed (alternatively as these determinations only occur in ``do_photometry`` then ``self.aperture_radius`` could be determined from ``image.shape`` rather than ``self.fitshape`` in that instance).

photutils/psf/tests/test_photometry.py
Onoddil: That does make more sense; I moved the function inside the corresponding test.

photutils/psf/tests/test_photometry.py
Onoddil: Added appropriate and (hopefully) unique ``match`` keywords to the ``warns``, to avoid this potential confusion.

Onoddil:
Changes have been made as requested @eteq - please review when possible.

eteq:
#903 somewhat collided with this as it turns out, but this provides a somewhat better solution for some cases.  So in an out-of-band discussion, @Onoddil, @larrybradley, and I think the way to move forward is to adjust it so that this PR's solution is adopted for models where `fitshape` exists, but for those where it does *not*, keep the behavior of #903.  That maintains the "helpful error message" aspect of #903, while also giving a reasonable fallback (with warning) for cases like epsf or fittableimagemodel where there's a reasonable guess available.

eteq:
Oops that last comment was slightly off - the solution I was proposing was for PSF models that are ePSF models (or possibly all `FittableImageModels`?), not just those with a `fitshape` (since the `fitshape` is set by the photometry object not the model)

Onoddil:
I have updated this PR to rebase onto current master and update based on previous discussions. While the previous PR #903 is an acceptable fallback condition, I do still prefer a less harsh fallback, so am in favour of this PR overriding the previous fallback condition.

While returning an error to the user to instruct them to add the parameter I think that the `fitshape` fallback is, for most users, a valid default. I have, however, updated the warning raised while doing so to point out to the user that they may want to consider if their `fitshape` _is_ appropriate, and to perhaps supply an `aperture_radius` anyway. This is fine because the `aperture_radius` is only used to calculate `flux_0`, an initial, order-of-magnitude starting flux to be passed to the main fitter, and so it doesn't have to be particularly precise.

The PR is currently blocked by #942 as I split out the very minor part of the original PR which was unrelated to this code change but is necessary for the `IterativelySubtractedPSFPhotometry` test in the tests to pass, so once that is merged I will rebase again and then the CI can be checked (I have verified it on my machine with the two lines back in, but you can't be too careful of course). I pushed these changes now so that they can be checked and discussed sooner rather than later.

I also wasn't 100% sure if we had done the 0.7 release as of yet so I've updated the changelog as though we haven't, since `CHANGES.rst` doesn't have a 0.8 changelog, but let me know if that is likely to change before the PR will be merged and I'll update it again.

cc @eteq @larrybradley for discussion of whether we still need the final `ValueError` or if the order of magnitude initial guess of `fitshape`'s aperture is good enough for most cases.

CHANGES.rst
```diff
@@ -126,8 +126,8 @@ Bug Fixes
   - Fixed a bug in the ``EPSFStar`` ``register_epsf`` and
     ``compute_residual_image`` computations. [#885]
 
-  - A ValueError is raised if ``aperture_radius`` is not input and
-    cannot be determined from the input ``psf_model``. [#903]
+  - Added fallback ``aperture_radius`` for PSF models without a FWHM or
+    sigma attribute, raising a warning. [#903, #740]
```
larrybradley: 0.7 was released yesterday.  The new changelog entry should go in 0.8 API changes.

photutils/psf/photometry.py
```diff
@@ -242,30 +242,21 @@ def do_photometry(self, image, init_guesses=None):
...
+            else:
+                self.aperture_radius = float(np.amin((np.asanyarray(
+                                             self.fitshape) - 1) / 2))
+                warnings.warn('aperture_radius is None and could not '
+                              'be determined by psf_model. Setting '
+                              'radius to the smallest fitshape size. '
+                              'If fitshape is significantly larger than '
+                              'the psf_model core lengthscale, consider '
+                              'supplying a specific aperture_radius.',
+                              AstropyUserWarning)
```
larrybradley: The logic here isn't correct.  The warning will be raised despite checking if the user inputed an `init_guesses` table with a `flux` column.  My deleted logic above is correct (see original lines 245-268).

photutils/psf/photometry.py
```diff
@@ -681,6 +672,21 @@ def do_photometry(self, image, init_guesses=None):
...
+                else:
+                    self.aperture_radius = float(np.amin((np.asanyarray(
+                                                 self.fitshape) - 1) / 2))
+                    warnings.warn('aperture_radius is None and could not '
+                                  'be determined by psf_model. Setting '
+                                  'radius to the smallest fitshape size. '
+                                  'If fitshape is significantly larger than '
+                                  'the psf_model core lengthscale, consider '
+                                  'supplying a specific aperture_radius.',
+                                  AstropyUserWarning)
```
larrybradley: This will need similar logic as in `BasicPSFPhotometry'.  Currently a warning will not be raised if `init_guess` is input without a `'flux'` column.  See block starting on new line 655.  All of this logic is the same and should not be repeated code, but extracted into a helper function if possible.

larrybradley: A *private* helper function.

larrybradley:
@Onoddil Since we already allow `aperture_radius` to be set by model FWHM, this is fine if `fitshape` can be determined.

CHANGES.rst
Onoddil: Done; I thought this would be an issue but I put the entry where it was supposed to go at the time. It's now in the 0.8 entry with the new release, and should hopefully now reflect an _update_ to 903 instead of replacing, but let me know if the wording isn't ok.

photutils/psf/photometry.py
eteq: To add just a bit more context - @Onoddil and I discussed that probably a larger effort in this direction is called for - I made #945 to track that. But I think @larrybradley is right that we may as well start on the right foot by doing it for this specific thing, even though probably more of this is called for.

Onoddil: While above in `BasicPSFPhotometry` the logic was unsound, here it is ok I believe. For iterative PSF photometry you will need `aperture_radius` set here, as in `_do_photometry` any newly detected sources will fail with the same error when trying to calculate the `aperture_flux` [at these lines](https://github.com/astropy/photutils/blob/692cdb17ed702ccd214526e0623bae7b6975fb59/photutils/psf/photometry.py#L739-L742).

These two piece of code are therefore subtly different and at this point I'm not sure you could wrap them into a helper function. Perhaps down the road once there are more iterative PSF photometry algorithms in place `BasicPSFPhotometry` can hold -- as the base class -- a helper function for _those_ cases, which would basically be this code moved out into its own function, but the iterative vs 'basic' cases here are different and should be standalone, I think.

I'm not sure how this links in with the larger issue @eteq, but I think I agree overall this function eventually should become a helper function, but `BasicPSFPhotometry` will remain an edge case which is handled specifically in its `do_photometry` call.

photutils/psf/photometry.py
eteq: I'm confused here.  I think the intent is that the warning should *not* be raised if init_guesses is present and has a flux column.  And that's true here, isn't it?  I.e. `if init_guesses is not None and 'flux_0' not in init_guesses.colnames` captures "the user input an init_guesses but it doesn't have flux"?, but if the flux is there the warning is *not* raised.  Or do you mean you want the *error* to be raised if init_guesses is None ?

Onoddil: Thanks for catching this, I did not remember to put that caveat back in. I've added this criterion back in to the `else` part of this to check for this again, as you had previously.

@eteq if the user supplies `init_guesses` which contains `flux_0` then we don't need to fallback on a default aperture radius; it's only if they give `init_guesses` _without_ `flux_0` that we need to worry. `if init_guesses is not None and 'flux_0' not in init_guesses.colnames` is @larrybradley's 'deleted logic' and is what is needed, yes.

Onoddil: Perhaps, thinking about it, I should add some kind of test to consider these eventualities...?

eteq: Yes, indeed, tests for each of these cases is a very good idea! (no need to worry about the actual results, so just a one-row table or something is fine).

photutils/psf/photometry.py
eteq: Ah, I missed the subtle difference here.  In that case I think it does make sense that they should be left separate for now.  

There are some ways that one could still imagine generalizing more usefully, like having the parts *inside* the `ifs` here be wrapped up in a private function with a few parameters (and be shared in basic and iterativelysubtracted), but I think it's better to do that in a follow-on because it will probably touch a lot of places and we wouldn't want it to drag this one out further.

photutils/psf/photometry.py
larrybradley: There are still issues here.  If `aperture_radius` is `None` and the PSF model doesn't have a FWHM or sigma attribute and:

1. `init_guesses` is `None`

or

2. `init_guesses` is input without a `flux_0` column

or

3. `init_guesses` is input with a `flux_0` column *and* the PSF model doesn't have a `fitshape` attribute

then in each of these three cases the code will not set `aperture_radius` and no informative error message will be raised -- thus the code will fail exactly like it did before I added the nice error message.

We really need to write tests for each of the above cases.

Onoddil: I agree that the first issue is _currently_ not covered as of the latest commit; that was a rushed job on my part to push the changes out before I went away and is my mistake, I blindly followed your suggested code instead of the _intent_ behind your comment, and forgot to add the second criterion in, which would of course solve `init_guesses is None`, which I can add; thanks for spotting the missing code. 

I'm not sure I follow on 2 and 3, however: 2 is covered by the code I did add to the `elif` catch after checking for `sigma` or `FWHM` attributes, isn't it? The docstring demands that `init_guesses` be an `astropy.table.Table` so checking for `flux_0 not in init_guesses.colnames` _does_ cover option 2, surely? All I did here was use the same criterion you did, so if yours worked I can only assume mine would too...

Similarly for 3, `fitshape` is the _routine's_ attribute, not the model (`self.fitshape` vs `self.psf_model.fitshape`), and cannot _not_ be input, with the docstrings requiring `int or length-2 array-like`; indeed, allowing `fitshape=None` as an input into `BasicPSFPhotometry` et al. is on my to do list, and thus (for now, at least) `self.fitshape` is _always_ acceptable (or the code falls over on the user before this point).

All of this is actually irrelevant, and my mistake for not realising and explaining my previous generic `else` catch more fully earlier: the next [block](https://github.com/astropy/photutils/blob/b9ba45925bcbe3d269e5f52f28282a59ef24f8e2/photutils/psf/photometry.py#L265-L299) of code just below this bit in `BasicPSFPhotometry` covers both of these eventualities: the first part covers `if init_guesses is not None and flux_0 not in init_guesses.colnames` while the `else` is therefore equivalent to `if init_guesses is None`. These two parts of the code block handle these issues, but simply require `aperture_radius` to be set, so we can just _always_ set `aperture_radius` if it isn't provided (and warn the user, of course), because code _already exists_ to account for these two issues.


I therefore actually would prefer, for the sake of clarity and code simplicity, to revert this minor change and go back to a generic `else` catch (with some comments -- either in the code or to the user -- to inform of why this is fine) after failing to find `sigma` or `FWHM` attributes. I will make these tests to ensure that these criteria behave as expected, of course, but with the code block already in place to handle the exceptions I think there shouldn't be any issues. If the fitting routine cannot find a `sigma` or `FWHM` attribute in `self.psf_model` then it falls back on an `aperture_radius` which is (hopefully) sensible (and gives the user to the warning to input their own, better one if it isn't), and uses that in the two cases we're worrying about, just as it would have done if the user had supplied their own `aperture_radius` anyway.

This actually then ties nicely into #945, as with that reverting to a generic `else` catch the code block for testing for the need of a fallback `aperture_radius` is _now_ in agreement between `BasicPSFPhotometry` and `IterativelySubtractedPSFPhotometry` and could be split off into a private helper function as originally desired; it is the next code block in `BasicPSFPhotometry` that handles the additional edge case interactions with `init_guesses` which the `_do_photometry` part of `IterativelySubtractedPSFPhotometry` doesn't have to deal with, since it always has to do aperture photometry for its runs beyond the first and so is forced to have an `aperture_radius` (handling its setup equivalent to `init_guesses is None` all the time).

larrybradley: Simplifying the code with the private helper function sounds great to me!  Thanks.

Onoddil:
I have pulled out the fallback aperture radius check into a helper function in `BasicPSFPhotometry` and applied it to both that and `IterativelySubtractedPSFPhotometry`. This should make more obvious the intent to set `aperture_radius` and let the next codeblock handle what happens if `init_guesses` isn't supplied, or is supplied without `flux_0` in `BasicPSFPhotometry.do_photometry`.

Note: this PR currently fails because I accidentally uncovered a bug in `IterativelySubtractedPSFPhotometry`, so it will require a rebase onto a subsequent PR to ensure the test passes before it can also be accepted.

Onoddil:
I have rebased this branch on to #986 and it now passes.

Waiting for review again, having resolved the above issues (hopefully!) - cc @eteq @larrybradley

larrybradley (approved):
Thanks, @Onoddil.

PR #745, opened 2018-10-09
Added the ability to pass additional column names to PSFPhotometry classes to be returned to the user

### LINKED ISSUE #680: PSFPhotometry: roundness and sharpness not in return table
Hi there, I was wondering if it is possible to retrieve the sharpness and roundness (1 and 2) from the sources detected by the DAOPhotPSFPhotometry class. The output table doesnt contain them, and i couldnt see a way of extracting it from your source code. These values are going to be of some importance to my results so i am looking for a way of getting them.
(I could match the data from DAOStarFinder but this would give me only the first iteration)
Thanks in advance!

Onoddil:
``output_cols`` is added as an additional input parameter to ``BasicPSFPhotometry``, ``IterativelySubtractedPSFPhotometry`` and ``DAOPhotPSFPhotometry`` allowing for the inclusion of a list of strings of parameters derived by the ``finder`` instance. If not ``None`` these additional columns are then passed back to the user in the output table. 

For example, if sharpness and roundness are parameters of interest from ``DAOStarFinder`` then ``output_cols = ['sharpness', 'roundness1', 'roundness2']`` would return the values calculated during the finder process, along with such values as ``x_fit``, etc.

Closes #680

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,7 +15,7 @@ API changes
 
 - ``photutils.psf``
 
-  - Added ``output_cols`` as a parameter to ``BasicPSFPhotometry``,
+  - Added ``extra_output_cols`` as a parameter to ``BasicPSFPhotometry``,
     ``IterativelySubtractedPSFPhotometry`` and ``DAOPhotPSFPhotometry``. [#745]
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,11 @@ Bug Fixes
 API changes
 ^^^^^^^^^^^
 
+- ``photutils.psf``
+
+  - Added ``output_cols`` as a parameter to ``BasicPSFPhotometry``,
+    ``IterativelySubtractedPSFPhotometry`` and ``DAOPhotPSFPhotometry``. [#745]
+
 
 0.7 (2019-08-14)
 ----------------
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -97,8 +97,9 @@ class BasicPSFPhotometry:
         keyword.  For analytical PSF models, alternatively you may
         define a FWHM attribute on your input psf_model.
     extra_output_cols : list of str, optional
-        List of additional columns for parameters derived by any of the intermediate
-        fitting steps (e.g., ``finder``), such as roundness or sharpness.
+        List of additional columns for parameters derived by any of the
+        intermediate fitting steps (e.g., ``finder``), such as roundness or
+        sharpness.
 
     Notes
     -----
@@ -225,8 +226,8 @@ class BasicPSFPhotometry:
             the initial guess for any parameters of the ``psf_model``
             model that are not fixed. If ``init_guesses`` supplied with
             ``extra_output_cols`` the initial values are used; if the columns
-            specified in ``extra_output_cols`` are not given in ``init_guesses``
-            then NaNs will be returned.
+            specified in ``extra_output_cols`` are not given in
+            ``init_guesses`` then NaNs will be returned.
 
         Returns
         -------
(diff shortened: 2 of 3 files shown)


larrybradley:
@Onoddil Thanks for the PR.

Why do you keep closing and reopening issues multiple times?  Please try not to do that.  An email is sent to all of the package watchers every time you close or (re-)open an issue.

johnnygreco:
Is the plan to still merge these changes? I'd love to use this option! Thanks!

photutils/psf/photometry.py
```diff
@@ -275,10 +280,21 @@ def do_photometry(self, image, init_guesses=None):
...
+                                         data=[sources['xcentroid'],
+                                               sources['ycentroid'],
+                                               sources['aperture_flux']])
+                else:
+                    init_guesses = Table(names=['x_0', 'y_0', 'flux_0'] +
+                                         list(self._output_cols),
+                                         data=[sources['xcentroid'],
+                                               sources['ycentroid'],
+                                               sources['aperture_flux']] +
+                                         [sources[o] for o in self._output_cols])
```
eteq: instead of the `if`/`else` I wonder if this is better done as leaving the original implementation and follow it with something like:
```
if self._output_cols is not None:
    for col_nm in self._output_cols:
        init_guesses[col_nm] = sources[col_nm]
    init_guesses[o] for o 
```

eteq:
@Onoddil 

> I cannot see where this point links to, so I am unsure of what the "line added here" is in reference to. Can you point to where this code would be added?

Oops, sorry, that was ambiguous.  I meant the line that I left the inline comment on (https://github.com/astropy/photutils/pull/745/files#diff-41d70bb987a66cf3476b4e99c1ad6c66R297).  But the idea is that you could do basically the same thing anywhere else in the whole process the table gets passed through (replacing `sources` in my snippet with the output of the later stages), because it would then only pick out columns that that stage produces (that's the point of the `if o in sources.colnames`).  Then it could really just be `output_cols` and the user doesn't have to worry exactly whether it came from the finder stage or fitter or some other stage we add later.

So if that seems practical, I favor 2, as I'm concerned about proliferation/over-specificity of keywords if we have to do e.g. `finder_output_cols`.

Onoddil:
Thanks @eteq, I think I have hopefully changed the code in a way roughly similar to how you had it in mind - I'm still slightly confused by the suggestions across two comments (and the inline comment seems to end slightly before the end of the recommended code change to me...).

Fundamentally the suggested way to add new columns to the original required ``Table`` was a good one but I've moved it to a convenience function in ``BasicPSFPhotometry`` to save a few lines and tidy the code up a bit, and this way it can be applied to any step, not just ``Finder``.

Someone may want to check if the way I've handled ``init_guesses`` and not passing all (or any) of the columns through in the first place (in which case I currently add the column and set all sources to ``np.nan`` for those columns not passed) is the correct/most pythonic/in line with assumed practises way to do things, and let me know how I should change that if it's not right.

jeffcarlin:
Will these changes be merged? This functionality could be super helpful, instead of having to match the DAOStarFinder and IterativelySubtractedPSFPhotometry outputs after the fact...

eteq:
I'm 👍 to @astrofrog's suggestion of changing the name to `extra_output_cols` here.

eteq:
So after some more thought, I see a potential issue that probably shouldn't block this PR but we may want to fix in a follow-on: As it turns out the extra cols currently *only* apply to the finder, because the stages are inconsistent about whether they pass everything on or not. So longer-term, we probably want to make that more consistent where every stage has some that are necessary and some parts that are optional.  Then we're back to the 2 vs 3 problem.  But I think the best choice at this point, since this PR clearly has a present need *right now*, is to do this as it stands, and then whenever we make the stages more consistent, we can make it so that `extra_output_cols` is *optionally* a dict (i.e., to allow for the "column names are used twice and therefore conflict" problem).  But that can be done later, since currently it's just the finder, so conflicts will never happen.

So I think the only change necessary for merging this now is to change the name to `extra_output_cols` and rebasing to fix the conflicts.  Then once this is merged I'll make a separate issue about the above disucssion.

Onoddil:
PR is back rebased onto latest master with all above comments taken on board: updated coverage test, new variable name, some extra PEP8 and/or deprecation 0.8 release things tidied up, and should be good for another review @eteq

CHANGES.rst
```diff
@@ -13,6 +13,11 @@ Bug Fixes
 API changes
 ^^^^^^^^^^^
 
+- ``photutils.psf``
+
+  - Added ``output_cols`` as a parameter to ``BasicPSFPhotometry``,
```
larrybradley:
```suggestion
  - Added ``extra_output_cols`` as a parameter to ``BasicPSFPhotometry``,
```

Onoddil: Quite right, sorry! Fixed.

eteq:
Alright, all good and tests are passing, so merging this.  Thanks @Onoddil !

eteq:
@johnnygreco @jeffcarlin - note this is now merged, so if you want to take it for a spin you can grab the latest developer version, or wait for the next release (0.8 I believe)

PR #746, opened 2018-10-09
Added flag to BasicPSFPhotometry to preserve ID order

### LINKED ISSUE #645: Sources order in BasicPSFPhotometry
When using `BasicPSFPhotometry` with Fixed Centroids the returned list of sources do not follow the same order as the input `init_guesses`, but rather the grouping order (thanks to the `hstack` at the end of the function), this could lead to confusion. 

I would suggest returning the results using the same order, or to place a clear warning in the documentation and/or docstring.

Onoddil:
Flag ``preserve_id_order`` added to ``BasicPSFPhotometry``, allowing for the re-sorting of ``output_table`` by source ID, overloading the default case where sources are returned in ascending ``group_id``. This would allow for sources to be returned in the same order as an input ``init_guesses``. Closes #645

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,7 +21,11 @@ Bug Fixes
 API changes
 ^^^^^^^^^^^
 
--
+- ``photutils.psf``
+
+  - Added ``preserve_id_order`` as input parameter to ``BasicPSFPhotometry``,
+    allowing for the return of source table in ``id`` rather than ``group_id``
+    order. [#746]
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


hamogu:
Is this really worth a new parameter? It's easy to reorder the final list on id. Alternatively, the return could always be ordered the same way as the input (instead of the current way where it's ordered by group id). 
We have so many parameters in the photometry already, please avoid adding more parameters where it's not really needed. More parameters mean more testing, more maintenance and more time for users until they understand what's going on.
Even #645 suggested that it might be sufficient to mention in the docs that stuff gets reordered in the process.

astrofrog:
I agree with @hamogu that it doesn't necessarily make sense to add a new parameter - I think it would make sense to instead always return in the same order as the input. Given that it isn't right now, I don't think anyone should be relying on the order of the output in some way, so it's probably safe to change the default behavior? (the opposite would probably not have been true - if we were changing away from ordering by id and to group_id then it might break people's code if they were assuming the output was in the same order as the input).

TL;DR I'd suggest changing the default order to be the same as the input and not having a kwarg

larrybradley:
The new PSF photometry tools automatically preserve ID order.

PR #750, opened 2018-10-16
Add filtering/selection by peakmax and brightest

mcara:
This PR adds two new parameters to `IRAFStarFind` and `DAOStarFind` star detection classes to allow source selection/filtering by:

1. `peakmax` - maximum peak pixel value for selecting stars.
2. `brightest` - number of brightest objects to keep.

The rationale for these parameters is that `peakmax` could allow automatic exclusion of sources that reach a known saturation level of the detector and `brightest` would keep top `brightest` stars eliminating the faintest objects that could be false detections.

CC: @larrybradley

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,12 @@ API changes
   - The ``find_peaks`` function will no longer issue a RuntimeWarning
     if the input data contains NaNs. [#712]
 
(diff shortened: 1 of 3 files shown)


larrybradley:
Thanks, @mcara.

Issue #757, opened 2018-10-29
Inconsistent tables returned by DAOStarFinder and IRAFStarFinder


mcara:
Currently, `DAOStarFinder` and `IRAFStarFinder` says that it returns table containing specific columns. For example, see [this link](https://github.com/astropy/photutils/blob/88d9fbab16fa2d1c59b6529a58ec88c6f66cc720/photutils/detection/findstars.py#L889) for `DAOStarFinder`. However, when there is a problem or no stars are found, instead of returning an empty table with specified columns, these functions return a really very very empty table.

This leads to issues such as https://github.com/spacetelescope/jwst/issues/2716 or, to look directly at the code: https://github.com/spacetelescope/jwst/blob/d7daa6c9cc60fc9870add2950ca06a424056db28/jwst/tweakreg/tweakreg_catalog.py#L59

Specifically, code that relies on the presence of advertised columns will crash. In my opinion, these two classes should be consistent in the format of returned tables.

mcara:
@larrybradley It was fixed in https://github.com/spacetelescope/jwst/pull/2796/files#diff-5224745bd38e8a6b9acd45aadc743da6R61 (awaiting approval).

mcara:
> I don't see any point in returning an empty table containing specific columns.

As long as it agrees with the documentation.

larrybradley:
One would need to check the length of the output table in any case (except I didn't do it in `tweakreg_catalog`), so adding column names to an empty table adds nothing.

mcara:
> adding column names to an empty table adds nothing

Not totally true: it adds information about expected headers. For example, when writing catalogs to files, see, e.g., 
https://github.com/spacetelescope/jwst/blob/d7daa6c9cc60fc9870add2950ca06a424056db28/jwst/tweakreg/tweakreg_step.py#L84, the headers is quite informative instead of getting an empty file (or not saving a file at all when the table is empty - that would be concerning).

eteq:
@larrybradley - I think I agree with @mcara that it makes sense to return an empty table with the columns.  I would think many users would do things like:
```python
...
newtable = table[table['xcentroid']<200]

... do stuff with newtable...
```
and it's useful for them to get an empty table if `table` has a length of 0.  So I think I agree with @mcara's PR here.

larrybradley:
@eteq Yes, I've been discussing this and other deep philosophical thoughts on the meaning of emptiness offline with @mcara and he's convinced me.  😄

[...]

PR #758, opened 2018-10-29
Make table output of DAOStarFinder and IRAFStarFinder consistent

mcara:
This PR make output of `DAOStarFinder` and `IRAFStarFinder` consistent regardless of whether any stars have been found. This addresses the issue https://github.com/astropy/photutils/issues/757

CC: @larrybradley


larrybradley:
Thanks, @mcara.

PR #760, opened 2018-10-29
Note that an empty table is returned if no sources are found

larrybradley:
Addresses #757.

CC: @mcara

### DIFF
--- a/photutils/detection/findstars.py
+++ b/photutils/detection/findstars.py
@@ -713,7 +713,8 @@ class StarFinderBase(metaclass=_ABCMetaAndInheritDocstrings):
         Returns
         -------
         table : `~astropy.table.Table`
-            A table of found stars.
+            A table of found stars.  If no stars are found then an empty
+            table is returned.
         """
 
         raise NotImplementedError('Needs to be implemented in a subclass.')
@@ -905,6 +906,8 @@ class DAOStarFinder(StarFinderBase):
             * ``mag``: the object instrumental magnitude calculated as
               ``-2.5 * log10(flux)``.  The derivation matches that of
               `DAOFIND`_ if ``sky`` is 0.0.
+
+            If no stars are found then an empty table is returned.
 
         .. _DAOFIND: http://stsdas.stsci.edu/cgi-bin/gethelp.cgi?daofind
         """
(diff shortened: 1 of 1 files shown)


photutils/detection/findstars.py
```diff
@@ -906,6 +907,8 @@ def find_stars(self, data):
               ``-2.5 * log10(flux)``.  The derivation matches that of
               `DAOFIND`_ if ``sky`` is 0.0.
 
+            If no stars are found then an empty table is returned.
```
mcara: I think the critical thing here is that we can have many kinds of "empty" tables: A table that has no rows but still has valid column headers or a table that has nothing - not even column names. IMO it would be better to return the first kind of "empty" tables.

mcara: This line does not specify the kind of emptiness that code is going to return

larrybradley: @mcara "kind of emptiness" sounds very philosophical! :smile:

mcara: The kind that `findstars` currently returns is "spooky"

mcara: What I am trying to say is that there is a phase transition between an empty _table_ and an empty _catalog_: a catalog has a structure (i.e., provided by column names).

mcara:
Here is a practical example when this might be useful: Let's say I am collecting all sources found in several images into one larger catalog. Then I can simply `vstack` all catalogs without first checking if I've got an empty table.

In case you wonder,

```python
In [14]: vstack([Table(names=['a','b']), Table(names=['a','b'])])
Out[14]: 
<Table length=0>
   a       b   
float64 float64
------- -------

In [15]: vstack([Table(names=['a','b']), Table()])
Out[15]: 
<Table masked=True length=0>
   a       b   
float64 float64
------- -------
```

For unknown to me reasons, in the second case I get a masked table???? (I guess this goes to a new `astropy.tables` issue...

mcara:
@larrybradley Actually, it seems that getting masked tables is [by design](http://docs.astropy.org/en/stable/api/astropy.table.vstack.html#astropy.table.vstack): 

> A value of ‘outer’ (default) means the output will have the union of all columns, **with table values being masked where no common values are available**.

Since an empty table does not have any values in common or not in common, maybe this is to be expected...?

larrybradley:
@mcara  Hmm, so the "empty" columns are masked (but aren't really part of the table).  Deep!

PR #756, opened 2018-10-26
WIP: Single Object Model and Scene Maker Future PSF Photometry API

Onoddil:
This pull request is set up for the discussion of the forward-looking additions to PSF photometry and should not be merged until after the discussion is finished. In this case the issue is the 'single object model' and 'scene maker' aspects of the [block diagram](https://github.com/astropy/photutils/blob/master/docs/psf_spec/block_diagram.png) -- although 'scene maker' is not currently implemented within the block diagram it is heavily linked to the 'single object model' and included here for discussion accordingly. This API documentation complements that in #721 where the documentation for those blocks implemented, with solid APIs unlikely to change in future releases. Please see #766 for an overview of the fitting process and a revised block diagram.

The structure of the blocks input and output parameters is described in detail in the documentation, but the main issues are summarised below:

- [ ] SingleObjectModel
  - [ ] Formatting
  - [ ] Wording
  - [ ] Internal model parameters
  - [ ] ``star`` variable renaming
  - [ ] Verification of internal consistency between ``object_type`` list available to both ``SingleObjectModel`` and ``SceneMaker``
  - [ ] API workflow: requires ``psf_model`` (`~astropy.modeling.Fittable2DModel`) and ``object_type`` (string) and returns ``convolve_psf_model`` (`~astropy.modeling.Fittable2DModel`)

- [ ] SceneMaker
  - [ ] Appropriate level of description of non-implemented API block
  - [ ] Wording
  - [ ] Formatting
  - [ ] Formalism for keeping track of merged sources

In particular discussion on 

- How to determine the 'internal' parameters of an extended source -- galaxy distance, time since supernova maximum, etc. -- within the framework of a ``Fitter`` instance. While point sources can be naively fit for flux and position with no knowledge of the underlying physics of the source (age, spectral type, etc.), the creation of an underlying intrinsic "image" of extended sources requires fitting for physical parameters.
- Whether it is necessary to rename the catalogue created during the PSF fitting process, currently called ``star``. Is it better to keep backwards consistency and keep all sources, extended or not, under a variable called 'star', or is it better to have a physically motivated name for the variable?
- What methods would be most effective in ensuring that ``SingleObjectModel`` and ``SceneMaker`` have the same accepted ``object_type`` list (such that ``SceneMaker`` cannot assign a source as being, e.g., a Quasar, which is not an accepted object type in ``SingleObjectModel``) -- or whether there is a more effective way to handle such non-point source determination than the list of accepted strings of ``object_type`` variables.
[...]

PR #755, opened 2018-10-25
WIP: Data Noise Handling Future PSF Photometry API

Onoddil:
This pull request is set up for the discussion of the forward-looking additions to PSF photometry and should not be merged until after the discussion is finished. In this case the issue is the 'data noise' (previously the 'noise_model'; the name of the block has been changed to avoid potential confusion with 'psf_model') aspect of the [block diagram](https://github.com/astropy/photutils/blob/master/docs/psf_spec/block_diagram.png). This API documentation complements that in #721 where the documentation for those blocks implemented, with solid APIs unlikely to change in future releases. Please see #766 for an overview of the fitting process and a revised block diagram.

The structure of the blocks input and output parameters is described in detail in the documentation, but the main issues are summarised below:

- [ ] NoiseData
  - [ ] Name
  - [ ] Formatting
  - [ ] Wording
  - [ ] Callable function vs more internal routine
  - [ ] Overload inputs keywords/function acceptable
  - [ ] Handling of conversion from uncertainty to ``weight`` keyword in ``fitter``
  - [ ] API workflow: requires ``data`` (array-like) and returns ``uncertainty`` (array-like)
  - [ ] Implications for ``aperture``

In particular discussion on 

- whether to revert the name back to 'noise_model', if this is too easily confused with 'psf_model', or if a third name is a better descriptor
- whether``noise_calc`` should be a simple ``callable`` function or a more internal routine of some kind 
- whether keyword overload (e.g., ``None`` allows for the simple effective unit-weighting of all data values and some secondary keyword, as yet unimplemented, to indicate that uncertainty values have been passed in with the data values and to ignore the computation of uncertainty values) is acceptable, and what such keywords should be
- the particular conversion of "uncertainty" to the ``weight`` keyword of ``fitter`` functions (i.e., is ``weight = 1/uncertainty``, are ``uncertainty`` values standard deviations or variance, etc)
- the implications of this API block -- primarily for PSF fitting -- on ``aperture`` fitting. Current implementation requires the update of ``aperture_photometry`` to accept the ``uncertainty`` argument (instead of its current ``error`` keyword) to correctly handle ``NDData`` arrays via ``@support_nddata``. Nominally a relatively minor and cosmetic keyword update, is it acceptable to force an API change on another part of `photutils` or must uncertainties be handled in a different way in the PSF fitting routines to avoid this conflict?

is appreciated to finalise the API call.

[...]

PR #753, opened 2018-10-19
WIP: Culler and Ender Future PSF Photometry API

Onoddil:
This pull request is set up for the discussion of the forward-looking additions to PSF photometry and should not be merged until after the discussion is finished. In this case the issue is the 'culler and ender' aspect of the [block diagram](https://github.com/astropy/photutils/blob/master/docs/psf_spec/block_diagram.png). This API documentation complements that in #721 where the documentation for those blocks implemented, with solid APIs unlikely to change in future releases. Please see #766 for an overview of the fitting process and a revised block diagram.

The structure of the blocks input and output parameters is described in detail in the documentation, but the main three stages are summarised below:

- [ ] ``CullerAndEnder`` 
  - [ ] Formatting
  - [ ] Wording
  - [ ] API workflow: requires ``data`` (`~astropy.table.Table`), ``psf_model`` (`~astropy.modeling.Fittable2DModel`), and ``new_sources`` (`~astropy.table.Table`) and returns ``new_data`` (`~astropy.table.Table`) and ``end_flag`` (boolean).
- [ ] ``cull_data``
  - [ ] Formatting
  - [ ] Wording
  - [ ] API workflow: requires ``data`` and ``psf_model`` and returns ``new_data`` (`~astropy.table.Table`)
- [ ] ``end_loop``
  - [ ] Formatting
  - [ ] Wording
  - [ ] API workflow: requires ``new_data``, ``data`` and ``new_sources``, returning ``end_flag`` (boolean)

Please provide any feedback on the API for this PSF photometry fitting routine block you may have, such that the implementation meets all of the requirements of all users and is as clear as possible going forward. A simple example for this block, maintaining "do nothing" backwards-compatibility can be found [here](https://github.com/Onoddil/photutils/blob/culler_ender_spec/photutils/psf/funcs.py), with [example](https://github.com/Onoddil/photutils/blob/culler_ender_spec/photutils/psf/photometry.py) implementation in ``IterativelySubtractedPSFPhotometry``. Please also provide simpler formatting or grammatical changes to improve the readability and professional look of the document. The function's primary goals are to remove poor quality fits from the list of sources returned to the user, and break the iterative process if certain criteria are met.

PR #721, opened 2018-09-24
WIP: PSF Photometry API documentation for established functionality

Onoddil:
This PR is part of the process for establishing the block API for PSF photometry. The documentation in this PR is unlikely to change as the four blocks here -- background estimation, source finder, source grouper, and PSF model -- are either independently used within `photutils` and thus well tested, matured and established pieces of code, or are relatively straightforward and thus leave little room for API changes. Please see #766 for an overview of the fitting process.

The documentation for ``background_estimator`` and ``finder`` should be very stable to changes, as these are the external function calls to routines used significantly throughout `photutils`. For these the primary issues are wording and formatting.

``group_maker`` however is slightly less stable, and thus in addition to checking for clarity and formatting this block description must succinctly and clearly explain the dependency of the group maker on the scene maker (developed in parallel with the extension of PSF fitting to non-point sources, in which "groups" can include the merging of several point sources into an extended object and vice versa).

``psf_model`` is a relatively minor documentation block, as it is primarily an external astropy class for which the PSF model must be a subclass, with specific input parameters. Thus this documentation must also explain this dependency.

The issues to be discussed in this PR are therefore:
- [ ] Background estimator
  - [ ] Formatting
  - [ ] Wording
- [ ] Finder
  - [ ] Formatting
  - [ ] Wording
- [ ] Group Maker
  - [ ] Formatting
  - [ ] Wording
  - [ ] Scene maker dependency
- [ ] PSF Model
  - [ ] Formatting
  - [ ] Wording
  - [ ] Fittable2DModel dependency

Please provide comments on any clarifications, inconsistencies, formatting changes or documentation flow changes here.

PR #766, opened 2018-11-06
WIP: Overview Documentation for PSF Fitting Workflow

### LINKED ISSUE #761: Make psf block diagram something that's actually maintainable
Right now the psf block diagram is basically a screenshot from keynote.  This is obviously un-maintainable in the face of any future changes.  Would be better to instead have it be, say, a `graphviz` diagram (which is easy to integrate into sphinx).

Onoddil:
This pull request is set up for the discussion of the overall PSF fitting process documentation, the fitting workflow and how each block interacts internally, and should not be merged until after the discussion is finished. This documentation complements that in #721 where the documentation for those blocks implemented, with solid APIs unlikely to change in future releases, and the newer APIs available in PRs #753, #755 and #756 for which details may be subject to change.

The main issues for discussion in this document are mostly limited to formatting and clarity of wording, as discussion of individual blocks should be directed to their respective PRs. The issues are summarised below:

- [ ] High Level Overview
  - [ ] Name
  - [ ] Wording
  - [ ] Formatting
  - [ ] Workflow graphic representation

One issue that should be discussed is the final graphical representation of the workflow. The current graph is simply a reproduction of the [block diagram](https://github.com/astropy/photutils/blob/master/docs/psf_spec/block_diagram.png), but this PR is meant to supersede that (and closes #761), and so confirmation of the workflow represented here is merited.

Please provide any feedback on the description of the overall PSF fitting process you may have, such that the implementation meets all of the requirements of all users and is as clear as possible going forward. Attached is the final output of the [graph](https://user-images.githubusercontent.com/3527254/48089546-c3410480-e1d2-11e8-93ed-96e8ede31368.jpeg) for viewing, as currently ``graphviz`` is not viewable within this PR.

### DIFF
--- a/docs/psf_spec/high_level_overview.rst
+++ b/docs/psf_spec/high_level_overview.rst
@@ -1,6 +1,32 @@
 PSF Fitting
 ===========
 
+.. role::  raw-html(raw)
+    :format: html
+
+`photutils` offers a module implementing a series of PSF fitting algorithms. These
+fitting algorithms are broken down into a series of submodules, allowing the user
+to implement specific versions of each part of the fitting process, implementing
+a variety of fitting procedures best suited to the specific science cases. However,
+at its most basic, all choices share the same basic functionality: a 2D image is
+supplied and a `~astropy.table.Table` of detected sources, with positions and flux
+counts, is returned.
+
+This simplified workflow is expanded below for the case of the
+`~photutils.psf.IterativelySubtractedPSFPhotometry` class, building upon the simpler
+`~photutils.psf.BasicPSFPhotometry` algorithm. While there are additional blocks
+used within the workflow, the main `~photutils.psf.BasicPSFPhotometry` flow is
+``image`` :raw-html:`&rarr;` ``finder`` :raw-html:`&rarr;` ``group maker``
+:raw-html:`&rarr;` ``fitter`` :raw-html:`&rarr;` ``stars``. The extension to more
+complex fitting classes, such as `~photutils.psf.IterativelySubtractedPSFPhotometry`,
+have the flow ``image`` :raw-html:`&rarr;` ``finder`` :raw-html:`&rarr;` ``group maker``
+:raw-html:`&rarr;` ``fitter`` :raw-html:`&rarr;` ``culler and ender`` :raw-html:`&rarr;`
+``finder`` :raw-html:`&rarr;` ... :raw-html:`&rarr;` ``stars``.
+
+Each submodule is described briefly here, with context for how it fits into the overall
+PSF fitting framework, providing a top-down view of the process. Where available, links
+to more detailed descriptions of each block is also provided.
+
 .. graphviz::
 
     digraph PSFFitting {
@@ -21,7 +47,7 @@ PSF Fitting
 
         w [label = "wcs"];
         i [label = "input image", shape = box];
-        gs [label = "guess at stars"];
+        gs [label = "guess at stars", shape = box];
         p [label = "psf model"];
         sm [label = "scene maker"];
         st [label = "stars\n(astropy.table)", shape = box];
@@ -62,3 +88,157 @@ PSF Fitting
         {rank = same; p; f; n;}
         {rank = same; gs; g; sm;}
     }
+
+
+.. _Image:
+
+Input Image
+^^^^^^^^^^^
+
+The input image is the most fundamental part of the PSF fitting process, as it is the
+product on which all subsequent methodology is applied. The image must be a two-dimensional
+array of ``x`` and ``y`` coordinates with ``flux`` represented by array values. If a
+WCS -- contained within a FITS file, for instance -- is passed with the two-dimensional
+array then the position of sources can be given in sky coordinates, instead of pixel values.
+The given image is passed as an input to the chosen PSF fitting class.
+
+
+.. _Finder:
+
+Finder
+^^^^^^
+
+The finder is the first step in the PSF fitting process, as sources must be discovered in
+the image before any kind of fit can be applied to them. These accept the input ``image``
+and produce a `~astropy.table.Table` of detected sources, by a set of criteria internal to
+the given ``Finder`` (see, e.g., `~photutils.detection.DAOStarFinder` using ``roundness``
+and ``sharpness`` to determine if sources are point-like). If an initial set of detected
+sources is passed to the fitter as ``init_guesses`` then ``Finder`` is not run on the first
+pass of an iterative fitting class, instead using the provided positions and fluxes. See
+the :doc:`finder <finder>` documentation for more details.
+
+
+.. _Group Maker:
+
+Group Maker
+^^^^^^^^^^^
+
+The second block to run in the fitting process, the various group maker processes, such as
+`~photutils.psf.DAOGroup`, allow for the merging of astrometrically overlapping sources
+into a set of multiple stars. These stars must be fit simultaneously (see Fitter_) as a
+composite model, avoiding the potential issue where blended sources return incorrect fluxes,
+as affects aperture photometry. The block accepts the `~astropy.table.Table` output from
+Finder_ and returns a second `~astropy.table.Table` containing the same columns as the input
+as well as an additional column indicating the group number of each source. More information
+on this block can be found in :doc:`Group Maker <group_maker>`.
+
+
+.. _Fitter:
+
+Fitter
+^^^^^^
+Once a set of sources has been detected and grouped, each source must have its respective
+properties -- position and flux primarily -- determined. For this a ``Fitter`` instance is
+required; these are drawn from a separate class or instance of minimization routines, such as
+those given under the `~astropy.modeling` umbrella. The fitter must be a two-dimensional fitter,
+by nature of the image to be fit, accepted as an input to the instance. The class must then
+return the position of the source (or sources, if a composite model was supplied; see
+`Group Maker`_) and its total flux, as determined by the PSF model describing the telescope
+system -- optics, CCD, etc. -- that created the image. See `below <PSF Model_>`_ for more
+details on the PSF model, and the :doc:`fitter <fitter>` documentation for further
+information on the ``Fitter`` instance and its properties.
+
+
+.. _PSF Model:
+
+PSF Model
+^^^^^^^^^
+Perhaps the most important aspect of PSF fitting, the ``PSF Model`` describes the distribution
+of light falling on a given pixel from a source at some position with some total flux, fully
+describing the effects of telescope optics, quantized CCD pixels, etc. In this respect it is
+more properly described as a PRF than a PSF -- see the :doc:`documentation <psf_model>` for
+more details -- but we conflate the two terms here freely. Similar to the ``Fitter``, it is
+a separate callable class detailing a specific PSF response. This can be some analytical
+function, such as `~photutils.psf.IntegratedGaussianPRF`, or an empirical description of the
+PSF, such as that implemented by the effective PSF functionality (see :ref:`build-epsf` for
+details). The model must describe a two-dimensional distribution, and must return the total
+flux of the source and its sub-pixel position via the ``evaluate`` function, accepted by the
+Fitter_ when minimizing the fit between the model and the data.
+
+
+.. _noise:
+
+Noise Description
+^^^^^^^^^^^^^^^^^
+The ``noise`` block within the PSF fitting process is, in much the same way as the ``Fitter``
+block, a separate ``callable`` function mapping the relationship between the pixel values in
+the input ``image`` and the corresponding uncertainty -- in units of standard deviations -- of
+the pixel. It must therefore accept a two-dimensional array and return an array of the same
+shape as ``image``. Alternatively it can be overloaded to indicate that the uncertainty array
+was passed along with ``image`` via an `~astropy.nddata.NDData` instance. A more detailed
+description of this block is available :doc:`here <noise_data>`.
+
+
+.. _Background Estimator:
+
+Background Estimator
+^^^^^^^^^^^^^^^^^^^^
+The ``image`` on which source extraction and evaluation is to be performed is not necessarily
+assumed to have had any pre-processing applied to it, and thus it may be necessary to account
+for the counts of the image background when handling PSF fitting. This block handles
+the determination of the background levels, avoiding high count peaks to calculate the typical
+count of an otherwise empty pixel. The block must therefore accepts the two-dimensional ``image``
+and returns either a single value -- the background count of the entire image, as calculated
+by the criteria of the specific block implementation -- or a 2D array of values, allowing for
+the possibility of varying background counts across the image. More information on this block
+can be found :doc:`here <background_estimator>`.
+
+
+.. _Culler and Ender:
+
+Culler and Ender
+^^^^^^^^^^^^^^^^
+Only implemented in iterative fitting classes, part of the extension to
+`~photutils.psf.BasicPSFPhotometry`, this block is the final step of a single iteration. While
+a maximum number of iterations can be specified for, e.g., 
+`~photutils.psf.IterativelySubtractedPSFPhotometry`, this block -- specifically, the ender half
+-- assesses whether all sources have been found within the ``image``. If no new sources have
+been found within an iteration, fitting can be stopped prematurely, avoiding wasteful computational
+time. Similarly, the culler aspect of this block examines sources found within an iteration for
+quality; it should calculate some goodness-of-fit criterion and reject sources picked up by the
+``Finder`` but fall below a given threshold. This block therefore accepts the ``stars``
+output `~astropy.table.Table` and returns a new `~astropy.table.Table` with low quality sources
+-- such as cosmic ray hits present in the image -- removed as well as a boolean flag indicating
+whether the iterative fitting process has reached a conclusion and can be terminated. Please refer
+to the :doc:`documentation <culler_and_ender>` for further details.
+
+
+.. _Single Object Model:
+
+Single Object Model
+^^^^^^^^^^^^^^^^^^^
+An additional block sitting between the `PSF model`_ and the Fitter_, the ``Single Object Model``
+extends the fitting capability of the class from purely point-source objects to extended sources.
+This block, effectively, convolves the PSF of the given observation with a physical light
+distribution, producing the flux seen in each pixel by the system of such an extended source. This
+block therefore requires as its inputs the PSF model and the object type -- a string representing
+the type of source (point-like, spiral galaxy, etc.) to create the intrinsic light distribution
+for -- and returns a new `~astropy.modeling.Fittable2DModel` instance, the convolution of the PSF
+and the true source. Further specifics on this block, including how it will link to `Scene Maker`_
+can be found in the :doc:`documentation <single_object_model>`.
+
+
+.. _Scene Maker:
+
+Scene Maker
+^^^^^^^^^^^
+The final block within the scope of PSF fitting, the scene maker is also the last to be implemented.
+Currently no PSF fitting class extends its functionality to include this block, but it will
+eventually generalize and possibly replace the `Group Maker`_ as the method of merging and assigning
+non-independent detections within a given image. While current grouping is simply the determining
+of point sources with potentially overlapping flux, the *scene* maker will allow for the grouping
+of both point sources into an overlapping, simultaneous fit group but also into a single extended
+object. Thus each iteration will require a step in which multiple-point-source extended objects are
+evaluated for their separation, and newly detected point sources are evaluated for potential
+assignment as part of an extended source. The details of this block are available :doc:`here
+<scene_maker>`, and will be expanded upon as the API is finalized.
--- a/docs/psf_spec/high_level_overview.rst
+++ b/docs/psf_spec/high_level_overview.rst
@@ -4,28 +4,31 @@ PSF Fitting
 .. role::  raw-html(raw)
     :format: html
 
-`photutils` offers a module implementing a series of PSF fitting algorithms. These
-fitting algorithms are broken down into a series of submodules, allowing the user
+The `photutils.psf` sub-package implements a variety of PSF fitting algorithms. These
+fitting algorithms are broken down into a series of modules, allowing the user
 to implement specific versions of each part of the fitting process, implementing
 a variety of fitting procedures best suited to the specific science cases. However,
 at its most basic, all choices share the same basic functionality: a 2D image is
 supplied and a `~astropy.table.Table` of detected sources, with positions and flux
-counts, is returned.
+counts, is returned. Optionally, the user can also supply a `~astropy.table.Table` of
+previously detected source positions as an input, for cases such as forced photometry
+where sources have previously been found in an image and their positions known already.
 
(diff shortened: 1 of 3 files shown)


docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+            // required to allow finder and guess to combine into group maker
+            invis [shape = circle, width = .1, fixedsize = true, style = invis];
+        }
+
+        w [label = "wcs"];
+        i [label = "input image", shape = box];
+        gs [label = "guess at stars", shape = box];
+        p [label = "psf model"];
+        sm [label = "scene maker"];
+        st [label = "stars\n(astropy.table)", shape = box];
```
eteq: I wonder of "stars" here should be "objects" or something like that?  Presumably the scene maker at least *might* have the option to do other kinds of fitting?

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+complex fitting classes, such as `~photutils.psf.IterativelySubtractedPSFPhotometry`,
+have the flow ``image`` :raw-html:`&rarr;` ``finder`` :raw-html:`&rarr;` ``group maker``
+:raw-html:`&rarr;` ``fitter`` :raw-html:`&rarr;` ``culler and ender`` :raw-html:`&rarr;`
+``finder`` :raw-html:`&rarr;` ... :raw-html:`&rarr;` ``stars``.
+
+Each submodule is described briefly here, with context for how it fits into the overall
+PSF fitting framework, providing a top-down view of the process. Where available, links
+to more detailed descriptions of each block is also provided.
+
+.. graphviz::
```
eteq: In general this is great!  One thing about this, though: It might be helpful to have some lines that pass through the image.  That is, have lines from where the image "enters" the big box to the finder and the background estimator?

And another thing: maybe try to add a "legend" that shows what the lines mean?  E.g. "solid = required flow of data", "dashed = optional information flow" and "dotted = optional component" or something like that?

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+        {rank = same; gs; g; sm;}
+    }
+
+
+.. _Image:
+
+Input Image
+^^^^^^^^^^^
+
+The input image is the most fundamental part of the PSF fitting process, as it is the
```
eteq: Should probably say something here about `NDData` objects.  That is, the fundamental object we will want to use to pass around the images is probably `NDData`.  There can just be something then that says roughly "but there's a way in that could instead use raw arrays or FITS files that then get auto-converted to `NDData"?  Or that might be a detail that can instead be in the other documents.

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+array then the position of sources can be given in sky coordinates, instead of pixel values.
+The given image is passed as an input to the chosen PSF fitting class.
+
+
+.. _Finder:
+
+Finder
+^^^^^^
+
+The finder is the first step in the PSF fitting process, as sources must be discovered in
```
eteq: These all need to be implementation of the `StarFinderBase` class, right? (i.e., https://photutils.readthedocs.io/en/stable/api/photutils.StarFinderBase.html#photutils.StarFinderBase).  If so you should say that, and link to it (e.g. ``  `~photutils.detectionStarFinderBase`  ``)

eteq: More broadly: I wonder if the `StarFinderBase` should actually be `ObjectFinderBase` or something like that?  Not necessary for this document, but worth thinking a bit about.

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+the :doc:`finder <finder>` documentation for more details.
+
+
+.. _Group Maker:
+
+Group Maker
+^^^^^^^^^^^
+
+The second block to run in the fitting process, the various group maker processes, such as
+`~photutils.psf.DAOGroup`, allow for the merging of astrometrically overlapping sources
```
eteq:
```suggestion
 `~photutils.psf.DAOGroup`, allow for the merging of sources that are astrometrically nearby
```

"overlapping" is a bit unclear because the definition of "overlapping" somewhat depends on the algorithm.  I think "nearby" is better in that it's intentionally vague?

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+into a set of multiple stars. These stars must be fit simultaneously (see Fitter_) as a
+composite model, avoiding the potential issue where blended sources return incorrect fluxes,
```
eteq:
```suggestion
 composite model, making deblending of sources possible.
```
(contingent on the next line suggestion)

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+as affects aperture photometry. The block accepts the `~astropy.table.Table` output from
```
eteq:
```suggestion
The block accepts the `~astropy.table.Table` output from
```

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+as well as an additional column indicating the group number of each source. More information
+on this block can be found in :doc:`Group Maker <group_maker>`.
+
+
+.. _Fitter:
+
+Fitter
+^^^^^^
+Once a set of sources has been detected and grouped, each source must have its respective
+properties -- position and flux primarily -- determined. For this a ``Fitter`` instance is
```
eteq:
```suggestion
 properties -- always position and flux, but sometimes other properties -- determined. For this a ``Fitter`` instance is
```

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+required; these are drawn from a separate class or instance of minimization routines, such as
+those given under the `~astropy.modeling` umbrella. The fitter must be a two-dimensional fitter,
```
eteq:
```suggestion
 those implemented in `~astropy.modeling.fitting`. The fitter must be a two-dimensional fitter,
```

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+by nature of the image to be fit, accepted as an input to the instance. The class must then
+return the position of the source (or sources, if a composite model was supplied; see
```
eteq: I think this part on the exact inputs/outputs could be left out because it's actually a bit more complicated (i.e., it has to conform to the `astropy.modeling.fitting` fitter API).  So just saying "it has to take in the image and table and output the fitted model.  More detail given in ..." or something like that is sufficient.

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+system -- optics, CCD, etc. -- that created the image. See `below <PSF Model_>`_ for more
+details on the PSF model, and the :doc:`fitter <fitter>` documentation for further
+information on the ``Fitter`` instance and its properties.
+
+
+.. _PSF Model:
+
+PSF Model
+^^^^^^^^^
+Perhaps the most important aspect of PSF fitting, the ``PSF Model`` describes the distribution
```
eteq:
```suggestion
 Perhaps the most variable aspect of PSF fitting, the ``PSF Model`` describes the distribution
```
?

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+of light falling on a given pixel from a source at some position with some total flux, fully
+describing the effects of telescope optics, quantized CCD pixels, etc. In this respect it is
+more properly described as a PRF than a PSF -- see the :doc:`documentation <psf_model>` for
```
eteq: I think it would be good to also link to the "terminology" section here: https://photutils.readthedocs.io/en/latest/psf.html#terminology (that is, link using sphinx syntax to the point in the docs that that link goes to), since that discusses the PRF/PSF/etc thing.

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+more details -- but we conflate the two terms here freely. Similar to the ``Fitter``, it is
```
eteq:
```suggestion
 more details -- but we use a single term here for brevity. Similar to the ``Fitter``, it is
```
we should probably fix this at some point and use the terminology more consistently?  But that can be done later, probably.

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+a separate callable class detailing a specific PSF response. This can be some analytical
+function, such as `~photutils.psf.IntegratedGaussianPRF`, or an empirical description of the
+PSF, such as that implemented by the effective PSF functionality (see :ref:`build-epsf` for
+details). The model must describe a two-dimensional distribution, and must return the total
+flux of the source and its sub-pixel position via the ``evaluate`` function, accepted by the
+Fitter_ when minimizing the fit between the model and the data.
```
eteq: As above with the fitter, I think it's best to describe it as it has to follow the `astropy.modeling.models` API for 2D models, because there's a lot of "little things" there that may end up mattering a lot.

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+
+
+.. _noise:
+
+Noise Description
+^^^^^^^^^^^^^^^^^
+The ``noise`` block within the PSF fitting process is, in much the same way as the ``Fitter``
+block, a separate ``callable`` function mapping the relationship between the pixel values in
+the input ``image`` and the corresponding uncertainty -- in units of standard deviations -- of
```
eteq: I'm not sure I like the "in units of standard deviation part", but will discuss more in #755

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+
+
+.. _Background Estimator:
+
+Background Estimator
+^^^^^^^^^^^^^^^^^^^^
+The ``image`` on which source extraction and evaluation is to be performed is not necessarily
+assumed to have had any pre-processing applied to it, and thus it may be necessary to account
+for the counts of the image background when handling PSF fitting. This block handles
+the determination of the background levels, avoiding high count peaks to calculate the typical
```
eteq: what do you mean by "avoiding high count peaks"?  I think that might be algorithm-specific, so maybe better to just do this:
```suggestion
 the determination of the background levels, and should provide the typical
```
?

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+by the criteria of the specific block implementation -- or a 2D array of values, allowing for
+the possibility of varying background counts across the image. More information on this block
+can be found :doc:`here <background_estimator>`.
+
+
+.. _Culler and Ender:
+
+Culler and Ender
+^^^^^^^^^^^^^^^^
+Only implemented in iterative fitting classes, part of the extension to
```
eteq: I don't understand the "part of the extension to" part... partly because `BasicPSFPhotometry` isn't really defined here (see my general comment)

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+Single Object Model
+^^^^^^^^^^^^^^^^^^^
+An additional block sitting between the `PSF model`_ and the Fitter_, the ``Single Object Model``
+extends the fitting capability of the class from purely point-source objects to extended sources.
+This block, effectively, convolves the PSF of the given observation with a physical light
+distribution, producing the flux seen in each pixel by the system of such an extended source. This
+block therefore requires as its inputs the PSF model and the object type -- a string representing
+the type of source (point-like, spiral galaxy, etc.) to create the intrinsic light distribution
+for -- and returns a new `~astropy.modeling.Fittable2DModel` instance, the convolution of the PSF
+and the true source. Further specifics on this block, including how it will link to `Scene Maker`_
```
eteq:
```suggestion
 and the true source. Further specifics on this block, including how it connects to `Scene Maker`_
```

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+^^^^^^^^^^^
+The final block within the scope of PSF fitting, the scene maker is also the last to be implemented.
+Currently no PSF fitting class extends its functionality to include this block, but it will
+eventually generalize and possibly replace the `Group Maker`_ as the method of merging and assigning
+non-independent detections within a given image. While current grouping is simply the determining
+of point sources with potentially overlapping flux, the *scene* maker will allow for the grouping
+of both point sources into an overlapping, simultaneous fit group but also into a single extended
+object. Thus each iteration will require a step in which multiple-point-source extended objects are
+evaluated for their separation, and newly detected point sources are evaluated for potential
+assignment as part of an extended source. The details of this block are available :doc:`here
```
eteq: This may be better discussed in #756 ... but this discussion makes me realize that maybe the scene maker and single object model should be one thing?  Or even if separate classes, maybe they should be described together in one document?

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+``Finder`` but fall below a given threshold. This block therefore accepts the ``stars``
+output `~astropy.table.Table` and returns a new `~astropy.table.Table` with low quality sources
+-- such as cosmic ray hits present in the image -- removed as well as a boolean flag indicating
+whether the iterative fitting process has reached a conclusion and can be terminated. Please refer
+to the :doc:`documentation <culler_and_ender>` for further details.
+
+
+.. _Single Object Model:
+
+Single Object Model
```
eteq: This (along with the scene maker) are *optional*, right?  That is, if not present, it means just use the PSF model directly?  In that case you should say that in both descriptions.

docs/psf_spec/high_level_overview.rst
marthaboyer: I'll second this. I was also going to suggest a description of what the different lines mean.

docs/psf_spec/high_level_overview.rst
marthaboyer: I second this also

docs/psf_spec/high_level_overview.rst
```diff
@@ -0,0 +1,244 @@
...
+`~photutils.psf.BasicPSFPhotometry` algorithm. While there are additional blocks
+used within the workflow, the main `~photutils.psf.BasicPSFPhotometry` flow is
+``image`` :raw-html:`&rarr;` ``finder`` :raw-html:`&rarr;` ``group maker``
+:raw-html:`&rarr;` ``fitter`` :raw-html:`&rarr;` ``stars``. The extension to more
+complex fitting classes, such as `~photutils.psf.IterativelySubtractedPSFPhotometry`,
+have the flow ``image`` :raw-html:`&rarr;` ``finder`` :raw-html:`&rarr;` ``group maker``
+:raw-html:`&rarr;` ``fitter`` :raw-html:`&rarr;` ``culler and ender`` :raw-html:`&rarr;`
+``finder`` :raw-html:`&rarr;` ... :raw-html:`&rarr;` ``stars``.
+
+Each submodule is described briefly here, with context for how it fits into the overall
```
marthaboyer: Minor fix to Erik's suggestion: A brief summary of each block is described here...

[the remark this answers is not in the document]
marthaboyer: Minor fix to Erik's suggestion: A brief summary of each block is described here...

docs/psf_spec/high_level_overview.rst
marthaboyer: Since there are details on the content of the image here, I agree with Erik that NDData objects should be mentioned.

docs/psf_spec/high_level_overview.rst
marthaboyer: agreed. Or at the very least, define the acronyms.

marthaboyer:
@Onoddil This all looks fantastic to me! I agree with Eriks suggestions & don't really have much else to add.

docs/psf_spec/high_level_overview.rst
Onoddil: I've updated the above references of ``stars`` to ``sources`` as that doesn't particularly affect anything and can be future-proofed now, but I obviously can't change these names as they're hard-coded links for now. Where would be best to raise this as a separate issue? Is actually making an issue the right approach for this?

docs/psf_spec/high_level_overview.rst
Onoddil: I was trying to lead into this section with the idea that while the rest of the sections (getting good sources, grouping blended things, etc.) are important, *this* section has to be done right or there's no point doing any of the other stuff at all... So important is what I meant (instead of variable, as in "most likely to change for each user"?), but this is slightly qualitative and perhaps subjective language; none of the other sections have lead-in.

I think the question is more:
do we just start this section with ```The ``PSF model`` describes the distribution```?

docs/psf_spec/high_level_overview.rst
Onoddil: I've made the change to the language, but to the more overall point: I'm not sure we really *do* need to change the terminology. Literally nowhere in the `photutils` framework do we NOT do PRF fitting, by nature of fitting quantised images we can't do anything but. Thus so long as we make it clear that technically PSFs can be (and are) pedantically different and that we are actually doing PRF fitting but we're just gonna use the colloquialism if that's ok with everyone else I think we get away with it.

Instead of putting loads of work into rewriting all of our documentation to use the new and more proper language, we just have to put work into making sure at any necessary point we differentiate between the two and link to the terminology page (like above!).

docs/psf_spec/high_level_overview.rst
Onoddil: I've removed the subclause for now as it's not particularly necessary (this being a brief, high level overview and all) but we can discuss that further in the other PR and I'll make any changes finalised there in this branch too.

docs/psf_spec/high_level_overview.rst
Onoddil: They're basically "point + extended" sources extensions of group maker and the PSF model, so while linked slightly more obviously by the need to share a common ``object_type`` list and know what to do with each available choice in each instance, they're no more or less inter-connected than group maker and PSF model; it's just obvious what to do with those as the only potential object type is a point source, which is simple to handle.

We can discuss this further in the corresponding PR to get into the details, and, again, I'll make any changes decided upon in there in this branch too.

docs/psf_spec/high_level_overview.rst
Onoddil: I've added language to this effect in the text, but eventually I don't think they'll be entirely optional, in that I think the block will always have to be used, but it can have convenient "do nothing" defaults. I think it's tied to the ``object_type`` flag that will eventually have to be thrown around in the ``astropy.table.Table`` objects: if that (by default) is something like ``point source`` then ``Single Object Model`` simply returns the input PSF model (i.e., f * &#948; = f) and ``Scene Maker`` just returns the input table of objects (i.e., no input sources are an extended source split apart).

This *effectively* means they're optional, as the user can just ignore that the more complex functionality exists and default parameters unchanged give the same answer as if they weren't there, but I think that's easier than having ``if use_som: new_model = single_object_model(psf)`` or equivalent.

Onoddil:
Thanks @eteq and @marthaboyer for proofs! I've made most of these changes without comment as they're good suggestions, but I've replied to a few inline comments for further question.

With regards to the larger comment of how ``BasicPSFPhotometry`` interacts with ``IterativelySubtractedPSFPhotometry``, I'm not sure how this should be tackled. I don't like that we focus so much on a single iterative fitting algorithm (as of writing it's the only one available, but see #732 for a new one...). The concept is that a "Basic PSF fit" is (see the second paragraph of this document for the basic flow, where I thought I discussed this but perhaps not obviously enough): ``image`` -> ``finder`` (optionally) -> ``grouping`` -> ``fitting`` -> ``sources``. An "Iterative PSF fit" does those things, then goes to ``culler and ender``, potentially interacts with ``image`` again (``IterativelySubtractedPSFPhotometry``, by the name, subtracts sources found in one iteration and re-fits the residuals, e.g.), potentially interacts with ``guess at sources`` (#732 stacks all previously detected sources to be re-fit together, instead of fitting sources in residuals, so it swaps image editing for output table editing), then goes back to ``finder``... Thus one could *actually* view a "Basic PSF fit" as an "Iterative PSF fit" with ``culler and ender`` defined to never cull any sources but always end the 'loop'. 

However, as I imagine we can't really get rid of ``BasicPSFPhotometry`` (as much as I'd like to just define a "Basic PSF fit" as "pick your favourite iterative routine but don't worry, set ``Culler and Ender`` to this ``BasicCullerEnder`` and it'll not make any difference"), I think we could better factor the codebase to use as much of ``BasicPSFPhotometry`` in subsequent subclasses as possible to highlight that it's basically just one iteration of the code. Thus while a ``BasicFit`` is ``image`` -> ``finder`` (optionally) -> ``grouping`` -> ``fitting`` -> ``sources``, ``IterativePSFPhotometry`` is ``BasicFit`` -> ``culler and ender`` -> (maybe end or) -> ``mess with image`` -> ``BasicFit`` -> ... -> end -> ``sources``.

As currently written I don't think it's quite so simple but I don't think the refactor to define things as such is too difficult; ``BasicPSFPhotometry`` and iterative subclasses do the heavy lifting in ``do_photometry`` but if that were changed so ``BasicPSFPhotometry`` essentially has ``def do_photometry(self, ...): return do_single_loop(self, ...)`` then subclasses could overload ``do_photometry`` but still use the super ``do_single_loop`` function.

For this documentation (instead of me waxing philosophical about what I'd like to do to the code...): the attempt to explain how ``IterativePSFPhotometry`` builds off ``BasicPSFPhotometry`` is in paragraph two. I added the box around the basic PSF parts of the block diagram, but it's basically everything except the culler and ender. I also added the part about "messing with the image" as an optional block. I think (given that I've just reproduced someone else's original block diagram) that the "subtract" floating between fitter and culler and ender was ``IterativelySubtractedPSFPhotometry``'s "subtract fit sources from image and find more in the residuals", so we should actually probably more generalise that either way, as that isn't true for all algorithms.

Let me know if the improvements to the block diagram help, and then any suggestions for how to better explain that ``IterativePSFPhotometry`` is the entirety of ``BasicPSFPhotometry`` + ``Culler and Ender`` + ``Tweak Input Image`` as a single iteration.

Onoddil:
Hi @marthaboyer and @eteq - as per a conversation I had offline with @eteq I've pushed the changes I made in response to the comments that were straightforward, but have a few inline replies to comments for clarifications, and the more general query in the comment above. I would appreciate it if you could review the changes I already made, have a look at the queries I had to a few of the comments you had previously, and have a think about the larger issue in my reply above this one.

larrybradley:
The 2017-2018 PSF photometry API plans (including the additions in this PR) have been archived on the `photutils` wiki page:  https://github.com/astropy/photutils/wiki/PSF-photometry-plans-(2017%E2%80%902018)

PR #824, opened 2019-04-12
combine instances of models instead of classes

nden:
An upcoming PR in astropy.modeling will deprecate and subsequently remove the ability to combine model classes. The one case of model class combination in photutils was replaced by combining model instances.  
Please let me know if a Change log is needed.


larrybradley:
Thanks, @nden!  No changelog is needed for this.

PR #826, opened 2019-04-19
Improve performance of ellipse fitting

ibusko:
This comes from work discussed in issue #802. 

In a nutshell, when profiling the code, we found  that vectorization of the `geometry.to_polar` method caused a significant performance degradation. In this PR we modify `to_polar` so it can handle scalar input quantities using scalar code, but still keeping its vectorized habilities. 

Profiling analysis indicates that the code returned to its overal performance level, similar to we had before.

Due to its own nature, the ellipse fitting algorithm as originally coded is slow. We are still investigating other approaches that involve a more in-depth vectorization of the entire algorithm, or caching of intermediate data (or both). This PR is to be taken as a stopgap solution to a pressing problem.

We tested this using both the test code provided in #802, as well as the notebook tutorials in `photutils-datasets`.

### DIFF
--- a/photutils/isophote/geometry.py
+++ b/photutils/isophote/geometry.py
@@ -415,12 +415,13 @@ class EllipseGeometry:
         # that is incurred when using vectorized code.
         #
         # The split in two separate functions helps in
-        # the profiling analysis.
+        # the profiling analysis: most of the time is
+        # spent in the scalar function.
 
         if type(x) == type(1) or type(x) == type(1.0):
-            self._to_polar_scalar(x, y)
+            return self._to_polar_scalar(x, y)
         else:
-            self._to_polar_vectorized(x, y)
+            return self._to_polar_vectorized(x,y)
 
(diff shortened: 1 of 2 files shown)


larrybradley:
CC @moustakas

moustakas:
Thank you, @ibusko!

I tested the pip-installed `photutils/v0.6` against the branch in this PR (`0.7.dev3364`) and indeed found a a factor of ~5 speed-up.  Nice!

I also made sure that the ellipse-fitting results are identical (not just the intensity, all attributes of the `IsophoteList` object) and indeed they are -- please feel free to merge this PR and close #802.  (If I have any further suggestions for improving the performance I'll open a new issue.)

larrybradley:
Thanks, @ibusko!  And thanks, @moustakas for testing the PR!

larrybradley:
Changelog entry added here: https://github.com/astropy/photutils/commit/19d92c86d5d6f0ee85e88b0eacdac6377b578eb3

PR #833, opened 2019-04-24
Checks for NDData and data/subshape dimension length agreement in subtract_psf

### LINKED ISSUE #818: subtract_psf in PSF Photometry fails for fitshape equal in length to input data
``subtract_psf``, used in ``BasicPSFPhotometry.nstar`` (and thus all classes based off this class) will fail with ``ValueError: Can't add array. Small array too large.`` if ``fitshape`` has shape equal in at least one length to ``data.shape``.

This is due to ``subtract_psf`` using ``astropy.nddata.add_array`` if ``subshape is None`` evaluates false. ``add_array`` fails to check if sizes are greater than _or equal_; I plan to put in a pull request through astropy to fix this. However, for exactly equal ``fitshape`` and ``data`` shape ``if subshape is None`` could be amended to ``if subshape is None or np.all(i == j for i, j in zip(fitshape, data.shape))`` or equivalent.

Onoddil:
This PR fixes a few minor check/edge cases in ``subtract_psf``.

1) The docstrings say that ``data`` can be an ``astropy.nddata.NDData`` object, but the first line previously checked for ``data.ndim`` which would throw an ``AttributeError`` for an ``NDData`` object.

2) Currently the only check on whether the PSF can be subtracted directly from the image or if a cutout is required is ``if subshape is None``, but equally true would be the case where ``data`` and ``subshape`` are equal in both dimensions, which bypasses unnecessary code in ``extract_array`` and ``add_array``.

Closes #818 (at least the case where both dimension lengths are equal; for the final edge case of one equal one non-equal dimension length ``add_array`` would need amending)

### DIFF
--- a/photutils/psf/utils.py
+++ b/photutils/psf/utils.py
@@ -7,6 +7,7 @@ import numpy as np
 from astropy.table import Table
 from astropy.modeling import models
 from astropy.nddata.utils import add_array, extract_array
+from astropy.nddata import NDData
 
 
 __all__ = ['prepare_psf_model', 'get_grouped_psf_model', 'subtract_psf']
@@ -207,10 +208,13 @@ def subtract_psf(data, psf, posflux, subshape=None):
 
     Returns
     -------
-    subdata : same shape and type as ``data``
+    subbeddata : same shape and type as ``data``
         The image with the PSF subtracted
     """
 
+    # if data has been passed as an NDData object, extract the data portion
+    if isinstance(data, NDData):
+        data = data.data
     if data.ndim != 2:
         raise ValueError('{0}-d array not supported. Only 2-d arrays can be '
                          'passed to subtract_psf.'.format(data.ndim))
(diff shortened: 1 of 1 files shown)


photutils/psf/utils.py
```diff
@@ -207,10 +208,13 @@ def subtract_psf(data, psf, posflux, subshape=None):
 
     Returns
     -------
-    subdata : same shape and type as ``data``
+    subbeddata : same shape and type as ``data``
```
astrofrog: what about ``subbed_data`` or ``subtracted_data``?

photutils/psf/utils.py
```diff
@@ -226,13 +230,15 @@ def subtract_psf(data, psf, posflux, subshape=None):
...
     psf = psf.copy()
     xname, yname, fluxname = _extract_psf_fitting_names(psf)
     indices = np.indices(data.shape)
     subbeddata = data.copy()
 
-    if subshape is None:
+    # data shape being exactly the same as subshape in both dimensions is equivalent to subtracting
+    # from the entire image, with no need to call extract_array and add_array
+    if subshape is None or all(data.shape[ind] == subshape[ind] for ind in [0, 1]):
         indicies_reversed = indices[::-1]
```
astrofrog: Can we fix indicies -> indices?

larrybradley:
The PSF tools have been rewritten and the  `subtract_psf` function no longer exists.

PR #834, opened 2019-04-25
Expanded FittableImageModel's oversampling factor to allow differing oversampling in x and y

Onoddil:
This PR gives ``FittableImageModel`` the option to accept ``oversampling`` factors that give differing values in the x and y direction, as a tuple of ``(x_oversamp, y_oversamp)``. It maintains backwards compatibility by accepting a scalar, interpreted as providing the same oversampling factors for both x and y (i.e., treating it as ``(oversampling, oversampling)``).

cc @eteq @larrybradley for comments. Forms part of a wider branch change with #815, #816 and #817 (see [here](https://github.com/astropy/photutils/pull/815#issuecomment-486709606) for more details).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -78,8 +78,8 @@ API changes
 
 - ``photutils.psf``
 
-  - ``FittableImageModel`` and subclasses now allow for two-dimensional
-    ``oversampling`` factors, specifying both oversampling in x and y. [#834]
+  - ``FittableImageModel`` and subclasses now allow for different ``oversampling``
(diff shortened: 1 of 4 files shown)


Onoddil:
I've fixed the slightly ambiguous language in the changelog entry, thanks for the better language @larrybradley. Yes, I think with the re-ordering of how these PRs seem to have to be accepted, I will now update #815 to remove the `np.mean` call; and have already updated the branch to use an array instead of individual `x_oversamp` variables.

larrybradley:
Thanks, @Onoddil.

PR #874, opened 2019-06-19
add area_overlap method for apertures

griffin-h:
Apertures have a method `area` that gives the exact area of the aperture. This can be used, e.g., to calculate the [local background](https://photutils.readthedocs.io/en/stable/aperture.html#local-background-subtraction) for aperture photometry. However this is the wrong number to use if the aperture is on the edge of the image. We need a method that returns the *overlapping* area between the aperture mask and the data. I have attempted to implement that in `area_overlap`. I wasn't sure if this was the right place to put it, so feedback is welcome.

### DIFF
--- a/photutils/aperture/core.py
+++ b/photutils/aperture/core.py
@@ -340,7 +340,7 @@ class PixelAperture(Aperture):
 
         Returns
         -------
-        areas : array_like
+        areas : float or array_like
             The overlapping areas between the aperture masks and the data.
         """
 
--- a/photutils/aperture/core.py
+++ b/photutils/aperture/core.py
@@ -347,8 +347,7 @@ class PixelAperture(Aperture):
         masks = self.to_mask(method=method, subpixels=subpixels)
         if self.isscalar:
             masks = (masks,)
-        masks_overlap = [mask.multiply(np.ones_like(data)) for mask in masks]
-        areas = [mask_overlap.sum() for mask_overlap in masks_overlap]
+        areas = [mask.multiply(np.ones_like(data)).sum() for mask in masks]
         if self.isscalar:
             return areas[0]
         else:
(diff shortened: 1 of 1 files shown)


photutils/aperture/core.py
```diff
@@ -297,6 +297,62 @@ def to_mask(self, method='exact', subpixels=5):
...
+                  between 0 and 1.
+
+        subpixels : int, optional
+            For the ``'subpixel'`` method, resample pixels by this factor
+            in each dimension.  That is, each pixel is divided into
+            ``subpixels ** 2`` subpixels.
+
+        Returns
+        -------
+        areas : array_like
```
larrybradley, line 343:
```suggestion
        areas : float or array_like
```

photutils/aperture/core.py
```diff
@@ -297,6 +297,62 @@ def to_mask(self, method='exact', subpixels=5):
...
+            The overlapping areas between the aperture masks and the data.
+        """
+
+        masks = self.to_mask(method=method, subpixels=subpixels)
+        if self.isscalar:
+            masks = (masks,)
+        areas = [mask.multiply(np.ones_like(data)).sum() for mask in masks]
```
larrybradley, line 350:
```suggestion
        data = np.ones_like(data)
        areas = [mask.multiply(data).sum() for mask in masks]
```

This will create the array only once instead of once per loop.

larrybradley (changes requested):
@griffin-h Thanks and sorry for overlooking this.  Overall this looks good.  I only have a couple minor requests before merging.  Thanks!

griffin-h:
> @griffin-h Thanks and sorry for overlooking this. Overall this looks good. I only have a couple minor requests before merging. Thanks!

No problem! I committed both your suggestions.

larrybradley (approved):
Thanks, @griffin-h.

PR #889, opened 2019-06-29
Halved computational time of photutils.background.background_2d.data_coords

JeffreyWardman:
Halved computational time of photutils.background.background_2d.data_coords. in background_2d._calc_coordinates() by using itertools.combinations instead of itertools.product and then filtering for values exceeding the smallest pixel axis (if x_max != y_max).

background_2d._calc_coordinates() is costly, and not a useful function. I would suggest deprecating it.

### DIFF
--- a/photutils/background/background_2d.py
+++ b/photutils/background/background_2d.py
@@ -720,8 +720,11 @@ class Background2D:
 
         # the position coordinates used when calling an interpolator
         nx, ny = self.data.shape
+        # Create an array consisting of all combinations of pixels up to max(nx, ny)
+        # then filter based on if nx and ny are not equal to each other.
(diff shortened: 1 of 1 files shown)


larrybradley:
@JeffreyWardman Thanks for your submission.  Unfortunately, `itertools.combinations` generates only *unique* (x, y) pairs, but what is needed here is *all* possible (x, y) pairs (order matters), hence the use of `itertools.product` instead.  Your patch doesn't produce the same results and therefore the package tests fails (even after fixing your itertools import that is causing the current CI failures).  If you have other ideas, I'd be very happy to include a faster way to produce the same results!

The `_calc_coordinates` function is needed because it defines `yx` (as a column stack), needed to initialize an interpolator, and `data_coords` (product of all possible pairs), needed to call an interpolator.  The code would fail without this method.

PR #892, opened 2019-07-17
Addition of columns to photutils.isophote._isophote_list_to_table()

HSouch:
It seems sensible to include the intensity gradients and central values for the isophotes (with associated errors) in the created QTable.


larrybradley:
Thanks, @HSouch.  This seems fine with me.  @ibusko?

larrybradley:
@ibusko, does this look OK to you to add these columns?

ibusko:
@larrybradley yes, it sounds fine. I thought the gradient was already included, but if it's not there, it should be. It was in the old STSDAS version. The central value is also a good thing to have.

I can take care of that, but with somewhat lower priority right now.

larrybradley:
Changelog added in https://github.com/astropy/photutils/commit/1526a4df25bb949ecb22331bca2dad9f209cd103

Issue #901, opened 2019-07-26
disallowing all zero SegmentationImage seems prohibitively restrictive


astromancer:
I've been happily inheriting from `SegmentationImage` for some time now [here](https://github.com/astromancer/obstools/blob/c54cdb6aff4eea34c1c5ee247bdf9eee68c256c3/phot/segmentation.py#L861) (this code is in development for a new pipeline for the SALTICAM  instrument of SALT as part of my Phd). However, [this](https://github.com/astropy/photutils/commit/a845d83cbbc36d9236d8b2060bb3f1ed96b04386#diff-00e7811220b26f668b507adf3d53f191) recent update broke some functionality of my code, specifically the ability to initialize this class with an all zero array.  I feel it's useful to be able to do so and that disallowing this is too restrictive:  My use case is one where I initially create the empty segmentation image and subsequently add labels to it for detections of progressively fainter objects: [example here](https://github.com/astromancer/obstools/blob/c54cdb6aff4eea34c1c5ee247bdf9eee68c256c3/phot/segmentation.py#L136)
Is there a specific reason why all zero `SegmentationImage` is now being disallowed? Seems like this bit of code slipped in with the test for other "obviously wrong" inf / nan values.

astromancer:
#836

larrybradley:
@astromancer A check was added disallowing an all-zero `SegmentationImage` because in that case the `SegmentationImage` does not work properly -- several of its attributes and methods raise errors because an all-zero image was never intended.  For example this could happen in an automated pipeline where no sources were detected, resulting in an all-zero segmentation images, and thus downstream attributes/methods would raise errors with no obvious reason.  Now the code explicitly raises a `NoDetectionsWarning`.

Also note that subclassing `SegmentationImage` did not work properly prior to https://github.com/astropy/photutils/pull/916 if the subclass did not manage reseting cached properties.  `SegmentationImage` caches properties and those caches would not be be properly reset in subclasses if the internal data changed.  That was fixed in version 0.7.

PR #902, opened 2019-07-26
additions to ease customization of SegmentationImage. addresses #901

astromancer:
I've added some code to `SegmentationImage` that will make it easier for subclasses to change behaviour when setting the `data` attribute.  This adds the function `set_data` which can be over-written by subclasses without needing to touch the property definition boilerplate. This allows for more flexible extension - inherited classes can choose to not implement the checks on user input data should they choose to do so. eg:
```
from photutils.segmentation import SegmentationImage

class MySegm(SegmentationImage):
    """subclass with less strict data validation"""
    def set_data(self, value):
        """nans / inf illegal, but non-positive allowed"""
        print(self.__class__, 'setting data')
        if np.any(~np.isfinite(value)):
            raise ValueError('data must not contain any non-finite values '
                             '(e.g. NaN, inf)')

        if '_data' in self.__dict__:
            # needed only when data is reassigned, not on init
            self._reset_lazy_properties()

        self._data = value

    def add_blob(self):
        new_data = self._data.copy()
        new_data[5:10, 5:10] = 1
        self.data = new_data

print('init')
seg = MySegm(np.zeros((20, 20)))
print('add blob')
seg.add_blob()
```
outputs:
init
<class '__main__.MySegm'> setting data
add blob
<class '__main__.MySegm'> setting data

### DIFF
--- a/photutils/segmentation/core.py
+++ b/photutils/segmentation/core.py
@@ -4,6 +4,7 @@ This module provides classes for a segmentation image and a single
 segment within a segmentation image.
 """
 
+import inspect
 from copy import deepcopy
 
 from astropy.utils import lazyproperty, deprecated
@@ -16,6 +17,10 @@ __all__ = ['Segment', 'SegmentationImage']
 
 __doctest_requires__ = {('SegmentationImage', 'SegmentationImage.*'):
                         ['scipy']}
+
+
+def is_lazyproperty(_):
+    return isinstance(_, lazyproperty)
 
 
 class Segment:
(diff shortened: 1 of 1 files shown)


larrybradley:
@astromancer To subclass `SegmentationImage` with custom data, you simply need to override the `data` property and setter, e.g.:

```python
class MySegmentationImage(SegmentationImage):
    def __init__(self, data):
        super().__init__(data)
    
    @property
    def data(self):
        return self._data
    
    @data.setter
    def data(self, value):
        # add custom data validation here

        self.__dict__ = {}   # important:  this resets cached properties
        self._data = value
```

In this simple example, `data` could be anything.  You would need to add some checks in the data setter as needed (e.g. NaN, inf).  Note though that the attributes/methods can raise errors in some cases (e.g. data array of all zeros).

astromancer:
The main point of this was actually using a separate `set_data` method in the parent class to handle input data: This would avoid the need to retype the whole property definition (as you have done above) and allows you to just overwrite `set_data` in order to change the way the class handles incoming data.  This reduces the amount of code you have to type when you want to inherit this class and change its behaviour.  Perhaps my example didn't make that clear.  With `set_data` from this PR included, your example above simplifies to:

```python
class MySegmentationImage(SegmentationImage):
    def set_data(self, value):
        # add custom data validation here

        self.__dict__ = {}   # important:  this resets cached properties
        self._data = value
```

Issue #849, opened 2019-05-19
How to use EPSFModel as psf_model in IterativelySubtractedPSFPhotometry?


ysBach:
I am relatively new to the PSF photometry (have just finished reading some parts of Stetson 1987 paper) and have no experience in IRAF either (I tried but always failed in PSF...).

Following the [doc](https://photutils.readthedocs.io/en/stable/epsf.html), I succeeded in making the EPSFModel (I also tested it to a globular cluster image that I have). At the end of the doc, there is a comment that
> The EPSFModel object is a subclass of FittableImageModel, thus it can be used as a the PSF model for the PSF-fitting machinery in Photutils (i.e. BasicPSFPhotometry, IterativelySubtractedPSFPhotometry, or DAOPhotPSFPhotometry).

so I naively put ``epsf`` to ``DAOPhotPSFPhotometry`` as below: 

 
```python
from photutils import datasets
from photutils.datasets import make_noise_image
from photutils import find_peaks
from astropy.table import Table
from astropy.stats import sigma_clipped_stats
from astropy.nddata import NDData
from photutils.psf import extract_stars, IntegratedGaussianPRF
from photutils import EPSFBuilder
from photutils.psf import DAOPhotPSFPhotometry as DAOPSF


hdu = datasets.load_simulated_hst_star_image()
data = hdu.data
data +=  make_noise_image(data.shape, type='gaussian', mean=10.,
                          stddev=5., random_state=12345)
peaks_tbl = find_peaks(data, threshold=500.)
peaks_tbl['peak_value'].info.format = '%.8g'  # for consistent table output
stars_tbl = Table()
stars_tbl['x'] = peaks_tbl['x_peak']
stars_tbl['y'] = peaks_tbl['y_peak']
mean_val, median_val, std_val = sigma_clipped_stats(data, sigma=2.)
data -= median_val
nddata = NDData(data=data)

stars = extract_stars(nddata, stars_tbl, size=25)

epsf_builder = EPSFBuilder(oversampling=4, maxiters=3,
                           progress_bar=False)
epsf, fitted_stars = epsf_builder(stars)
photter = DAOPSF(crit_separation=4 * fwhm,
                 threshold=thresh,
                 fwhm=fwhm,
                 psf_model=epsf,
                 sharplo=0,
                 sharphi=20,
                 roundlo=-20,
                 roundhi=20,
                 fitshape=3)
photter(data)
```

but getting the `TypeError`:
```
---------------------------------------------------------------------------
[...]

Issue #804, opened 2019-03-10
EPSModel has no attribute fwhm or sigma


SunilSimha:
I'm using version 0.6 and there is no way I can use `IterativelySubtractedPSFPhotometry` because the `EPSModel` class has no attribute `fwhm`. When I tried using `IterativelySubtractedPSFPhotometry`, it gave an error that said:
```bash
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-81-0462a0dc654c> in <module>()
----> 1 result_tab=photometry(image=testimg)
      2 residual_image = photometry.get_residual_image()

~/anaconda3/lib/python3.6/site-packages/photutils/psf/photometry.py in __call__(self, image, init_guesses)
    185         """
    186 
--> 187         return self.do_photometry(image, init_guesses)
    188 
    189     def do_photometry(self, image, init_guesses=None):

~/anaconda3/lib/python3.6/site-packages/photutils/psf/photometry.py in do_photometry(self, image, init_guesses)
    660                                             gaussian_sigma_to_fwhm)
    661 
--> 662             output_table = self._do_photometry(['x_0', 'y_0', 'flux_0'])
    663         return output_table
    664 

~/anaconda3/lib/python3.6/site-packages/photutils/psf/photometry.py in _do_photometry(self, param_tab, n_start)
    701             apertures = CircularAperture((sources['xcentroid'],
    702                                           sources['ycentroid']),
--> 703                                          r=self.aperture_radius)
    704             sources['aperture_flux'] = aperture_photometry(
    705                 self._residual_image, apertures)['aperture_sum']

~/anaconda3/lib/python3.6/site-packages/photutils/aperture/circle.py in __init__(self, positions, r)
    122 
    123     def __init__(self, positions, r):
--> 124         if r < 0:
    125             raise ValueError('r must be non-negative')
    126 

TypeError: '<' not supported between instances of 'NoneType' and 'int'
```
Digging through the code revealed that r or `aperture_radius` wasn't defined properly in this block of code starting from line 235 of `photometry.py`

```python
if self.aperture_radius is None:
            if hasattr(self.psf_model, 'fwhm'):
                self.aperture_radius = self.psf_model.fwhm.value
            elif hasattr(self.psf_model, 'sigma'):
                self.aperture_radius = (self.psf_model.sigma.value *
                                        gaussian_sigma_to_fwhm)
```
[...]

PR #903, opened 2019-07-29
Raise ValueError in PSF fitting if aperture_radius is not set

### LINKED ISSUE #728: About PSF model

larrybradley:
`aperture_radius` is used to compute initial estimates for the fluxes of sources.  If initial flux guesses are not input to the photometry class via the ``init_guesses`` keyword, then `aperture_radius` must be set.  Currently, a cryptic error is raised if that's not the case.  This PR raises a `ValueError` with a helpful message if `aperture_radius` is needed and it is not set or cannot be determined from the input `psf_model`.

This issue was reported in #728, #804, and #849 (and via other channels).
Closes #728.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -105,6 +105,9 @@ Bug Fixes
   - Fixed a bug in the ``EPSFStar`` ``register_epsf`` and
     ``compute_residual_image`` computations. [#885]
 
+  - A ValueError is raised if ``aperture_radius`` is not input and
+    cannot be determined from the input ``psf_model``. [#903]
+
 - ``photutils.segmentation``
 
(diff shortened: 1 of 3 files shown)


larrybradley:
The `AppVeyor` failure is unrelated.

Issue #904, opened 2019-07-29
`make_source_mask` fails with AttributeError for all zero data


astromancer:
Related to recent API changes in #836.  Since `detect_sources` no longer always returns a `SegmentationImage`, the assumption in the [final line](https://github.com/astropy/photutils/blob/b079c3908eb436a531d31a02a8c252ee0c101cf1/photutils/segmentation/detect.py#L270) of `make_source_mask` that it does is no longer correct.

To reproduce:
```
from photutils.segmentation.detect import make_source_mask
make_source_mask(np.zeros((100, 100)), 3, 5)
```
Traceback:
```
/usr/local/lib/python3.6/dist-packages/astropy/utils/decorators.py in wrapper(*args, **kwargs)
    519                             kwargs[new_name[i]] = value
    520 
--> 521             return function(*args, **kwargs)
    522 
    523         return wrapper

/usr/local/lib/python3.6/dist-packages/photutils-0.7.dev3690-py3.6-linux-x86_64.egg/photutils/segmentation/detect.py in make_source_mask(data, snr, npixels, mask, mask_value, filter_fwhm, filter_size, filter_kernel, sigclip_sigma, sigclip_iters, dilate_size)
    268 
    269     selem = np.ones((dilate_size, dilate_size))
--> 270     return ndimage.binary_dilation(segm.data.astype(np.bool), selem)

AttributeError: 'NoneType' object has no attribute 'data'
```

larrybradley:
@astromancer Many thanks for reporting this bug!   It's fixed in #917.

PR #906, opened 2019-07-29
add detect_array function #904

astromancer:
This PR adds the `detect_array` function - source detection that always returns an array.  This fixes the erroneous `AttributeError` in #904. As a bonus, this also makes this package a bit more modular. This function is expected to be useful in other cases where it is sufficient to work with an array, and also precludes condition testing for `SegmentationImage` vs `None` as would be necessary for return values from `detect_sources`.

```
from photutils.segmentation.detect import make_source_mask
make_source_mask(np.zeros((100, 100)), 3, 5)
```
outputs
```
array([[False, False, False, ..., False, False, False],
       [False, False, False, ..., False, False, False],
       [False, False, False, ..., False, False, False],
       ...,
       [False, False, False, ..., False, False, False],
       [False, False, False, ..., False, False, False],
       [False, False, False, ..., False, False, False]])
```

### DIFF
--- a/photutils/segmentation/detect.py
+++ b/photutils/segmentation/detect.py
@@ -125,6 +125,21 @@ def detect_sources(data, threshold, npixels, filter_kernel=None,
         ax2.imshow(segm.data, origin='lower', interpolation='nearest')
     """
 
+    segm_img = detect_array(data, threshold, npixels, filter_kernel,
+                            connectivity, mask)
+    if np.sum(segm_img) == 0:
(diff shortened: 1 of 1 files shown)


larrybradley:
It's trivial to get an array instead of a `SegmentationImage`:

```python
segm = detect_sources(data, ....)
arr = segm.data if segm is not None else np.zeros(data.shape)
```

I'd rather not add a new function to the API to replicate this simple line.

The `make_source_mask` `AttributeError` from #904 was fixed in #917.

PR #907, opened 2019-07-31
Update orientation units in narrative docs

larrybradley:
Minor fix to narrative docs to ensure correct units for source `orientation`.


larrybradley:
`AppVeyor` failure is unrelated.

PR #912, opened 2019-08-02
Fix #911: linear mode not working

### LINKED ISSUE #911: linear=True is not implemented photutils.isophote.ellipse.fit_image()
I am trying to fit isophotes to an image with constant spacing between them using the procedure outlined here (https://photutils.readthedocs.io/en/stable/isophote.html). However, the linear keyword does not appear to be having any effect on the table that is outputted; regardless of how it is set, the same grid of sma are given for the step that I choose.

"
from photutils.isophote import EllipseGeometry
geometry = EllipseGeometry(x0=xbp_ref, y0=ybp_ref, sma=8, eps=0.3,
                           pa=100.*np.pi/180.)

from photutils import EllipticalAperture
aper = EllipticalAperture((geometry.x0, geometry.y0), geometry.sma,
                          geometry.sma*(1 - geometry.eps),
                          geometry.pa)

from photutils.isophote import Ellipse
ellipse = Ellipse(median_flux, geometry)

isolist = ellipse.fit_image(step=1, linear=True)

isolist.to_table()
"

I glanced at the source code and did not see a check for the linear keyword anywhere. Has this feature not been implemented yet?

ibusko:
This fixes the problem reported in issue #911

### DIFF
--- a/photutils/isophote/ellipse.py
+++ b/photutils/isophote/ellipse.py
@@ -203,7 +203,7 @@ class Ellipse:
                   conver=DEFAULT_CONVERGENCE, minit=DEFAULT_MINIT,
                   maxit=DEFAULT_MAXIT, fflag=DEFAULT_FFLAG,
                   maxgerr=DEFAULT_MAXGERR, sclip=3., nclip=0,
-                  integrmode=BILINEAR, linear=False, maxrit=None):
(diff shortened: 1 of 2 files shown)


larrybradley:
Thanks, @ibusko!

PR #919, opened 2019-08-07
Docstring tweaks in psf/epsf

eteq:
This does a couple of straightforward documentation tweaks mainly in the PSF and ePSF realm.  Specifically, mainly it's focusing on getting rid of "default is X" statements in the docstrings.

My reasoning here is that the default is already specified in what the default actually *is*, and both the `?` tool in ipython/jupyter and the sphinx docs display that pretty clearly.  So philisophically I think it's best not to state it in the docstring because that's asking for trouble if someone changes the default in the code but not the docstring.

cc @Onoddil @larrybradley

### DIFF
--- a/photutils/psf/models.py
+++ b/photutils/psf/models.py
@@ -497,9 +497,11 @@ class FittableImageModel(Fittable2DModel):
 
 class EPSFModel(FittableImageModel):
     """
-    A class that models an effective PSF (ePSF). This is a subclass of
-    `FittableImageModel`, and is very similar but contains a few additional
-    parameters relevant specifically for ePSFs.
+    A class that models an effective PSF (ePSF).
+
+    While this class is a subclass of `FittableImageModel`, it is very similar.
+    The primary differences/motivation are a few additional  parameters necesary
+    specifically for ePSFs.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


eteq:
The CI failure is unrelated I think (see #921 )

photutils/psf/models.py
```diff
@@ -497,8 +497,12 @@ def evaluate(self, x, y, flux, x_0, y_0, use_oversampling=True):
 
 class EPSFModel(FittableImageModel):
     """
-    A subclass of `FittableImageModel`. A fittable ePSF model.
+    A class that models an effective PSF (ePSF). This is a subclass of
+    `FittableImageModel`, and is very similar but contains a few additional
+    parameters relevant specifically for ePSF's.
```
Onoddil:
```suggestion
    parameters relevant specifically for ePSFs.
```

My only comment is a very minor grammar fix.

photutils/psf/models.py
```diff
@@ -497,8 +497,12 @@ def evaluate(self, x, y, flux, x_0, y_0, use_oversampling=True):
 
 class EPSFModel(FittableImageModel):
     """
-    A subclass of `FittableImageModel`. A fittable ePSF model.
+    A class that models an effective PSF (ePSF). This is a subclass of
```
larrybradley: Nit:  Split the second sentence into a new paragraph so there's a nice one-line summary to start the doctring.  (Unfortunately multi-line suggestions are not available in github yet).

larrybradley:
The test failure is unrelated.

PR #929, opened 2019-08-12
Fixes gradient computation.

ibusko:
This gives more stability to the fit algorithm on low SNR regions. But in some cases may result in slight differences in results.


bsipocz:
@ibusko - could you please rebase? There are some unrelated changes, and the tests seem to segfault, but to be sure about it would be great to be on top of master.

larrybradley:
I'm trying to run the tests locally.  The `test_ellipse.py` tests run significantly slower now (~130 sec *without* using the `--remote-data` option), and the `test_model.py` tests are effectively hung (running locally for > 10 min before I killed it).  The CI failures are due to timeouts from the `test_model.py` tests.

@ibusko If there a quick fix for this?  If not, then I'm going to release `0.7` without it.

ibusko:
I am working on the tests now.

The reason test_ellipse takes longer seems to be that it’s fitting lower SNR regions that where not seen by the former algorithm.

test_model seems to be running into a numerical precision situation, also caused by the fit going to regions where it didn’t go before.

On Aug 12, 2019, at 3:38 PM, Larry Bradley <notifications@github.com<mailto:notifications@github.com>> wrote:


I'm trying to run the tests locally. The test_ellipse.py tests run significantly slower now (~130 sec without using the --remote-data option), and the test_model.py tests are effectively hung (running locally for > 10 min before I killed it). The CI failures are due to timeouts from the test_model.py tests.

@ibusko<https://github.com/ibusko> If there a quick fix for this? If not, then I'm going to release 0.7 without it.

ibusko:
I noticed that my fork has something broken that I was unable to fix with a rebase. I will close this PR, start a new fork, and generate a new PR from a branch in there. That should remove the unwanted commits from the commit history.

PR #937, opened 2019-08-14
PSF fixes and code cleanup

larrybradley:
Mostly minor miscellaneous code fixes and cleanup in the psf subpackage.  Also, this PR adds some missing changelog entries from  #817 due to API changes and backwards-incompatible changes.

The one major fix is in https://github.com/astropy/photutils/commit/2c9cbdf55ad83f0a91864ccd8cde677880d9e974 which fixes the logic for defining the EPSF centroids.  The first `centroid_func` call was overwritten by a subsequent call, thus the `shift_val` keyword value had no effect.

CC: @Onoddil, @eteq, @bsipocz

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -261,6 +261,13 @@ API changes
   - Removed deprecated ``Star``, ``Stars``, and ``LinkedStar``
     classes. [#894]
 
+  - Removed ``recentering_boxsize`` and ``center_accuracy`` keywords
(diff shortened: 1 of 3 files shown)


Onoddil:
Hi @larrybradley, thanks for fixing the issues, and finding the bugs, in #817. One minor note is that your fix for the issue with `flux` being an unused variable in `EPSFModel._initial_norm` is rather that it lost its return statement; I've put in a minor PR #939 to tweak the fix for that one.

PR #939, opened 2019-08-14
Fix to EPSFModel initial_norm, where flux failed to be returned

Onoddil:
A very small pull request to tweak a bugfix for a bug found and fixed in #937, where in `EPSFModel._initial_norm` the `flux` assignment incorrectly failed to be returned, rather than being an unused variable (cf. the equivalent function in `FittableImageModel` [here](https://github.com/astropy/photutils/blob/59a85c05dcf4735da7a13bdcefb95c154e0a3600/photutils/psf/models.py#L162-L171)), as was implemented in the above PR.


larrybradley:
Thanks, @Onoddil!

PR #961, opened 2019-10-07
Fix a labeling bug in deblend_sources

### LINKED ISSUE #959: Extremely elongated apertures from deblended segmentation image
Extracting and deblending galaxy sources from a fairly noisy field, I have a curious problem, where some of the apertures generated as described in the [online documentation](https://photutils.readthedocs.io/en/stable/segmentation.html#image-segmentation) are extremely elongated, all being close to horizontal. The same does not happen when not deblending the sources, but does happen for a variety of deblending parameter settings and both linear and exponential modes. 

[Data file](https://github.com/astropy/photutils/files/3685501/F225Wcutout.zip)

[Jupyter Notebook reproducing the problem](https://github.com/astropy/photutils/files/3685508/Cutouts.zip)

larrybradley:
This PR fixes #959, which was caused by a re-labeling bug introduced in v0.7 in `deblend_sources`.

Many thanks to @thriveth for reporting the original bug and providing the data and a MWE that enabled me to debug this issue!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -84,6 +84,15 @@ API changes
     ``assert_angle``, and ``pixel_to_icrs_coords``. [#953]
 
 
+0.7.1 (unreleased)
+------------------
+
+Bug Fixes
+^^^^^^^^^
+
+- Fixed a labeling bug in ``deblend_sources``. [#961]
+
+
 0.7 (2019-08-14)
 ----------------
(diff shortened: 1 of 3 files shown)


larrybradley:
The travis-ci failure is due to upstream changes in `astropy` dev, which I will fix in a separate PR.

Issue #951, opened 2019-08-23
How to build ePSF with provided star fits


oxno2:
I tried to obtain the star PSF for a very large image of HST.  The original image was huge, I tried to obtain the stars first and put them into individual fits. After subtracted the background and masked some contamination sources, I wanted to use the EPSFBuilder to combine these processed stars images and generate an effect native PSF. There were nearby sources around the original stars so I mask these regions with numpy.nan.

I followed the same method from the example document. I read the values of the stars and stored in a numpy array called "stars_sci" and then combined the data with np.hstack. 
```
data=numpy.hstack(star_sci)

#Then I just follow the same code from the example document.

import matplotlib.pyplot as plt
from astropy.visualization import simple_norm
from photutils import datasets
from photutils.datasets import make_noise_image
peaks_tbl = find_peaks(data, threshold=5000)  
peaks_tbl['peak_value'].info.format = '%.8g'  # for consistent table output  
print(peaks_tbl)  
size = 25
hsize = (size - 1) / 2
x = peaks_tbl['x_peak']  
y = peaks_tbl['y_peak']  
mask = ((x > hsize) & (x < (data.shape[1] -1 - hsize)) &
            (y > hsize) & (y < (data.shape[0] -1 - hsize)))  
from astropy.table import Table
stars_tbl = Table()
stars_tbl['x'] = x[mask]  
stars_tbl['y'] = y[mask]  

from astropy.stats import sigma_clipped_stats
mean_val, median_val, std_val = sigma_clipped_stats(data, sigma=2.)  
data -= median_val  
#The extract_stars() function requires the input data as an NDData object. An NDData object is easy to create from our data array:

from astropy.nddata import NDData
nddata = NDData(data=data)  

from photutils.psf import extract_stars
stars = extract_stars(nddata, stars_tbl, size=25) 

```
The stars looked ok 
![DeepinScreenshot_select-area_20190822223031](https://user-images.githubusercontent.com/22302177/63564904-9032ec00-c52c-11e9-8707-b6cb57d1f547.png)
However, the final EPSF was very strange. I didn't know what happened and why the shape looked so irregular.
![DeepinScreenshot_select-area_20190822223055](https://user-images.githubusercontent.com/22302177/63564909-94f7a000-c52c-11e9-9148-eba6b3ec9b02.png)

cloud182:
[...]

Issue #860, opened 2019-06-07
Strange negative patterns in the ePSFs


gkacprzak:
Hello,

Thank you for creating this great package as it has been quite easy to use s far. I have been having some troubles understanding the hashed structure seen in the ePSFs. I find great PSFs (oversampling=1) for both space-based and ground-based images. However, as soon as I increase the oversampling, this grid pattern occurs and has this negative grid that appears and this effect increases with increasing the oversampling. It's harder to tell from the oversampling=5 ground based image but it is there. Can someone please help me understand this?  Thanks

<img width="870" alt="ePSFs" src="https://user-images.githubusercontent.com/14156119/59081757-25addf00-8933-11e9-87e3-740629234411.png">

ysBach:
Had similar issue when I was preparing the lecture note for this spring's TA note ([this note](https://nbviewer.jupyter.org/github/ysbach/AO2019/blob/master/Notebooks/08-PSF_Extraction.ipynb)):

![image](https://user-images.githubusercontent.com/17214293/61196089-4a9e2a80-a707-11e9-8d22-04ab5a462b88.png)
(In the note, I didn't strictly followed Stetson's original PSF extraction technique. Now I think I should've followed the Gaussian + spline technique, but anyway I extracted the PSF without Gaussian component, i.e., pure spline interpolation).

When I tested, smaller number of iteration gave more severe such patterns. Then I just increased the number to 100 and it got better. The "strange patterns" then appeared only at the outskirt of the PSF, so I thought maybe it's not a big problem. I don't quite understand the theory of ePSF, so sorry I can't give meaningful discussion right now...

larrybradley:
Ping: @Onoddil

eteq:
@gkacprzak   - we *think* #974 should have solved the problem.  Can you give the latest development build of `photutils` a try? (Or if you aren't set up to install from source, try it when 0.7.2 comes out.)  If that addresses your problems here, we can close this issue.

@ysBach - it's less clear to me if your problem is the same as what #974 fixed.  Can you give the latest `master` a try and hopefully that will determine if it's the same or a different issue?

larrybradley:
Closing this as fixed by #2419.

[...]

PR #974, opened 2019-10-21
Fix to ePSF algorithm changes in v0.7

### LINKED ISSUE #969: EPSF gets worse at each iteration
I am trying to build an EPSF with the ``EPSFBuilder`` class - however I've found that at each iteration the EPSF actually gets worse and worse:

Screenshot from 2019-10-18 12-49-01

Here is the full notebook:

https://gist.github.com/astrofrog/7b2c3a8c462547ffe9367f6ab28131ed

Could this indicate a bug in the EPSF building, or is the algorithm inherently unstable?

This occurs with the latest stable version (0.7.1) as well as 0.8.dev0

Onoddil:
This PR fixes the bug in `EPSFBuilder` in its v0.7 update. The bug, as best I pinned it down, is due to a subtle interaction between a "halfway" algorithm update to the original Anderson & King (2000) paper, from its form in v0.6 which was, as I understand it, implemented via discussions with Jay, unbeknownst to me at the time. I have subsequently found what I believe to be a citable version of the updates (and referenced them accordingly), and thus re-aligned the ePSF algorithm with its updated version. Ironically if I had made more code changes instead of fewer, and further aligned the code with the workflow of AK00 it would likely have been ok, but there we are...!

The main differences should therefore lie in the removal of the `residual smooth and recenter` loop of AK00, to follow [ISR WFC3 2016-12](http://www.stsci.edu/hst/wfc3/documents/ISRs/WFC3-2016-12.pdf); for continuity I simply reversed the recentering algorithm completely, keeping the original iteratively-find-the-center-via-centroiding function. There should hopefully be very few differences in the v0.6 and v0.7 results -- some preliminary testing on very poor datasets (16 randomly placed Gaussian stars with sigma of 0.5 pixels, noisy images and poor initial centroids for an oversampling factor of 4) gives results that are not complete gibberish as with the previous examples in issues raised, so hopefully this should successfully revert the bug.

This PR also implements the minor change to include a custom `SigmaClip`, allowing for custom sigma clip functions to be passed to `EPSFBuilder`.

This PR does _not_ include changes to one major difference between the v0.6 algorithm and both AK00 and ISR 2016-12: whether the residual sampling to grid point mapping is a nearest neighbour, "box of half a grid spacing length" function or, as is the case in the two cited works, residual samplings are placed in all grid points within a grid spacing (i.e., a box of a _full_ grid spacing length, 0.25 pixels of `oversampling=4`) for the purposes of sigma-clipping residuals for ePSF grid interpolation point updating. Given the current issue being one of non-obvious changes to cited works being made through offline, internal discussions I felt perhaps it was wise to hold off on this change before confirming whether this change was also a case of out-of-band 'unofficial' updates/suggestions. Is this the case, or should I make a separate PR to implement the swap from a one-to-one sampling:grid point mapping to a each-sampling-is-put-in-four-grid-points mapping? I can see arguments for reducing the distance a sampling point can be from any grid point, to reduce the effects of data points from 'extreme' distances affecting the sigma-clipped residual point.

Finally, this PR has come about through a fairly roundabout route, with several rebases a long the way, so it would be good to get a few pairs of eyes on the code to ensure no further bugs are introduced trying to fix the first bug. I believe the changes to at least be robust to _some_ level of my poking the code with a stick, but cc @eteq @larrybradley for some fresh perspective!

Fixes #969

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -86,16 +86,8 @@ Bug Fixes
 
 - ``photutils.psf``
 
-  - Fix to algorithm in ``EPSFBuilder``, causing issues where ePSFs
-    failed to build. [#974]
-
-API changes
-^^^^^^^^^^^
-
-- ``photutils.psf``
-
-  - Added ``flux_residual_sigclip`` as an input parameter, allowing for
-    custom sigma clipping options in ``EPSFBuilder``. [#974]
+  - Fix to algorithm in ``EPSFBuilder``, addressing issues where ePSFs
+    failed to build (yielding striped ePSFs). [#974]
 
 
 0.7.1 (2019-10-09)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -78,6 +78,26 @@ API changes
     ``assert_angle``, and ``pixel_to_icrs_coords``. [#953]
 
 
+0.7.2 (Unreleased)
+------------------
+
+Bug Fixes
+^^^^^^^^^
+
+- ``photutils.psf``
+
+  - Fix to algorithm in ``EPSFBuilder``, causing issues where ePSFs
+    failed to build. [#974]
+
+API changes
+^^^^^^^^^^^
+
+- ``photutils.psf``
+
+  - Added ``flux_residual_sigclip`` as an input parameter, allowing for
+    custom sigma clipping options in ``EPSFBuilder``. [#974]
+
+
 0.7.1 (2019-10-09)
 ------------------
 
--- a/docs/epsf.rst
+++ b/docs/epsf.rst
@@ -21,9 +21,11 @@ Building an ePSF
 
 Photutils provides tools for building an ePSF following the
 prescription of `Anderson and King (2000; PASP 112, 1360)
-<http://adsabs.harvard.edu/abs/2000PASP..112.1360A>`_ The process
-involves iterating between the ePSF itself and the stars used to build
-it.
+<http://adsabs.harvard.edu/abs/2000PASP..112.1360A>`_ and subsequent
+enhancements detailed mainly in
+`Anderson (2016), ISR WFC3 2016-12 <www.stsci.edu/hst/wfc3/documents/ISRs/WFC3-2016-12.pdf>`_.
+The process involves iterating between the ePSF itself and the stars
+used to build it.
 
 To begin, we must first define a sample of stars used to build the
 ePSF.  Ideally these stars should be bright (high S/N) and isolated to
--- a/photutils/psf/epsf.py
+++ b/photutils/psf/epsf.py
@@ -291,6 +291,18 @@ class EPSFBuilder:
         The undersampled value at which to compute the shifts.  It must
         be a strictly positive number.
 
+    recentering_boxsize : float or tuple of two floats, optional
+            The size (in pixels) of the box used to calculate the centroid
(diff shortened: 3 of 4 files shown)


Onoddil:
I also haven't included the changelog yet, holding off so I can get some opinions on what type of changes various things are, whether they go in 0.7.1 or 0.8, etc. I assume the bug fix of the `EPSFBuilder` algorithm is 0.7.1, and therefore the API change of `SigmaClip` goes along with it, but the bug fix also contains API changes (with the re-introduction of `center_accuracy` and `recentering_boxsize`, e.g.)

Finally, this reversal of changes throws into question the now-default `centroid_epsf` as a) the default centroiding algorithm, and b) useful at all, since it's just for `EPSFBuilder` currently. The re-inclusion of `box_size` in `_recenter_epsf` makes it a bit awkward to handle the AK00 centroiding function, as it just involves the few pixels either side of a given shift value. Opinions on whether that has been handled correctly in the code, and if we should change the default again, are appreciated.

Onoddil:
Attached is a screengrab of running the notebook in #969 with these changes, showing qualitatively the same PSF as shown in the issue for v0.6's algorithm.

check_issue_969_convergence

eteq:
There are several issues this should fix - @Onoddil pointed out #969 is fixed by this, and I just checked @larrybradley's notebook from https://github.com/astropy/photutils/issues/951#issuecomment-537980934 , with proof that this fix addresses that here:
image

Note I *believe* this should fix the other people who mentioned this in #951, as well as #860, but cannot test those.  I think we should plan on merging this since it fixes at last part of that, and then ask the folks there to test.

Onoddil:
I've made the changes as requested by @eteq -- changing `sigclip`'s name, and minor edits to the new test.

The one outstanding issue I have, which might require a quick update after feedback, is whether we should also roll back the default centroiding algorithm. Currently the change in v0.7 holds, where the default centroiding algorithm is `centroid_epsf`, the algorithm detailed in AK00; however, if this is a larger "roll back miscommunication re: changes to the algorithm since 2000" then perhaps the default centroid algorithm shouldn't be the one laid out in the 2000 paper?

The centre-of-mass algorithm is certainly more flexible as a default, and I _do_ use it in the new testing (although offline tests show that the symmetry centroiding of `centroid_epsf` also works), so should I put the default back to `centroid_com`?

eteq:
After some out-of-band discussion with @larrybradley, the conclusion was that it is indeed better to add the `flux_residual_sigclip` in 0.8.  But so as to make this easy to merge quickly, I've just pushed up a few commits "hiding" the functionality to be merged in this PR for 0.7.2, and then #984 brings it back (which will then be milestoned 0.8).

So hopefully if the tests are happy this is now good to go (but should probably get a glance from @larrybradley before merging)

emirkmo:
I don't know if this is the right place to write this but 0.72 still gives me a diverging PSF fit instead of converging to a good fit when using EPSFBuilder. I.e., in 0.71 and 0.72 I get the same behavior as mentioned in: #https://github.com/astropy/photutils/issues/969

All versions >= 0.7 give the same diverging result. However, 0.6X versions correctly converge. So for now, I had to downgrade to that. This also means Astropy needs to be downgraded to a version before 4.0.

Let me know if an example would help. In my tests, the issue was not solved.

PR #984, opened 2019-11-14
Restore flux_residual_sigclip

eteq:
This was implemented by @Onoddil in #974 but because it was a new feature we decided to "hide" it for 0.7.2.  This PR just brings it back for 0.8.

Note I'm starting it as a "draft" because it needs to be rebased on master after #974 is in.


larrybradley:
@eteq Can you please rebase this?

PR #986, opened 2019-11-14
Fix to IterativelySubtractedPSFPhotometry failing when Finder does not return None if no sources are detected

Onoddil:
Currently the `_do_photometry` loop of `IterativelySubtractedPSFPhotometry` assumes that if no sources are detected then the returned variable will be `None`. If a user provides a custom finder which instead returns an empty `Table` object, this won't be caught by the while loop and will error. This PR makes a simple fix to this, adding a check for a zero-length variable returned by `self.finder` in addition to a `None` variable.

This PR blocks #740, which needs this PR merged before one of its tests will pass.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,12 @@ New Features
 Bug Fixes
 ^^^^^^^^^
 
+- ``photutils.psf``
+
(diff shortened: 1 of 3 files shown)


Onoddil:
Fair enough -- moved the changelog to `0.7.2` :)

PR #1008, opened 2019-12-06
Fix upper harmonics calculation

### LINKED ISSUE #1004: b4 parameter
Hi,
I am analysing the b4 parameter from the elliptical isophote fitting. I am getting very different (orders of magnitude higher) values than what I am getting from IRAF task 'ellipse' for very similar geometrical parameters. Does anyone know why does this happen ?

From Jedrzejewski's original paper, it seems that there is some transformation at play:

> To interpret the amplitudes obtained, they are transformed from intensity coefficients to coefficients describing deviations from a unit circle as in Carter (1977), by dividing by the local slope of the intensity profile and by the semi-major axis length.

So, I checked if this is the missing part, but apparently not.

ibusko:
This fixes an error in the upper harmonics calculation as explained in #1004. 

Fixes #1004


larrybradley:
@ibusko There's a real test failure in `test_isophote.py: TestIsophote.test_m51()` where `b4` is now `None`:
```
        # deviations from perfect ellipticity
        assert abs(iso.a3) <= 0.05
        assert abs(iso.b3) <= 0.05
>       assert abs(iso.a4) <= 0.05
E       TypeError: bad operand type for abs(): 'NoneType'
```
https://travis-ci.org/astropy/photutils/jobs/621651635#L3956

photutils/isophote/tests/test_isophote.py
```diff
@@ -59,11 +59,13 @@ def test_fit(self):
...
-        assert abs(iso.b3) <= 0.01
-        assert abs(iso.a4) <= 0.01
-        assert abs(iso.b4) <= 0.01
+        # deviations from perfect ellipticity. Note
+        # that sometimes a None covariance can be
+        # generated by scipy.optimize.leastsq
+        assert iso.a3 is None or abs(iso.a3) <= 0.01
+        assert iso.a3 is None or abs(iso.b3) <= 0.01
+        assert iso.a3 is None or abs(iso.a4) <= 0.01
+        assert iso.a3 is None or abs(iso.b4) <= 0.01
```
larrybradley: The first conditionals here all copy `iso.a3`, but should be `b3`, `a4`, `b4`.  The tests are still failing because of this.

photutils/isophote/tests/test_isophote.py
```diff
@@ -97,11 +99,13 @@ def test_m51(self):
...
-        assert abs(iso.b3) <= 0.05
-        assert abs(iso.a4) <= 0.05
-        assert abs(iso.b4) <= 0.05
+        # deviations from perfect ellipticity. Note
+        # that sometimes a None covariance can be
+        # generated by scipy.optimize.leastsq
+        assert iso.a3 is None or abs(iso.a3) <= 0.05
+        assert iso.a3 is None or abs(iso.b3) <= 0.05
+        assert iso.a3 is None or abs(iso.a4) <= 0.05
+        assert iso.a3 is None or abs(iso.b4) <= 0.05
```
larrybradley: Here too.

ibusko: Indeed, it's not the correct way to test for the None condition. But since the 4 parameters are all tied together, meaning, either none of them, or all of them, can have the value None, the test would work anyway. But would look funny to someone wanting to understand the logic.

ibusko: The next commit fixes it.

ibusko:
@larrybradley how do we fix the bot changelog complaint?

photutils/isophote/tests/test_isophote.py
larrybradley: Thanks, @ibusko.  Actually, this `Travis` test failed because `iso.a4` is `None`, but `a3` and `b3` apparently are not `None` (see https://travis-ci.org/astropy/photutils/jobs/621673519#L3958).

 In any case, I think your fix is clearer and more robust.

PR #1039, opened 2020-05-21
Break the while loop outside the image [Fixes #1038]

### LINKED ISSUE #1038: build_ellipse_model falls in to an infinite loop if object extends outside the image boundary
While preparing the lecture note, I realized the previously-worked ``build_ellipse_model`` part never stops.
An example (identical to the tutorial [here](https://photutils.readthedocs.io/en/stable/isophote.html), only increasing the size of the galaxy so that large portion of it extends outside the image):
```python
#%%
import numpy as np
from astropy.modeling.models import Gaussian2D
from photutils.datasets import make_noise_image
g = Gaussian2D(100., 75, 75, 50, 12, theta=40.*np.pi/180.)
ny = nx = 150
y, x = np.mgrid[0:ny, 0:nx]
noise = make_noise_image((ny, nx), distribution='gaussian', mean=0.,
                         stddev=2., random_state=12345)
data = g(x, y) + noise

#%%
from photutils.isophote import EllipseGeometry
geometry = EllipseGeometry(x0=75, y0=75, sma=20, eps=0.5,
                           pa=20.*np.pi/180.)

#%%
import matplotlib.pyplot as plt
from photutils import EllipticalAperture
aper = EllipticalAperture((geometry.x0, geometry.y0), geometry.sma,
                           geometry.sma*(1 - geometry.eps),
                           geometry.pa)
plt.imshow(data, origin='lower')
aper.plot(color='white')

#%%
from photutils.isophote import Ellipse
ellipse = Ellipse(data, geometry)
isolist = ellipse.fit_image()

#%%
from photutils.isophote import build_ellipse_model
model_image = build_ellipse_model(data.shape, isolist)
residual = data - model_image
```

I found there is a while loop in the function, which does not end if the pixel is outside the image. A simple addition of ``else: break`` solved the issue.

ysBach:
This PR fixes the issue #1038 (``build_ellipse_model``). 

Only two lines are added
```python
    else:
        break
```
and moved a one-line coment just above this ``else``. (Tests passed on my computer, just for confirmation)

Fix #1038

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,7 +20,7 @@ Bug Fixes
 - ``photutils.isophote``
 
   - Fix a bug that ``build_ellipse_model`` falls into an infinite loop
-    when the pixel to fit is outside of the image. [#1038]
+    when the pixel to fit is outside of the image. [#1039]
 
 - ``photutils.segmentation``
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,11 @@ New Features
 Bug Fixes
 ^^^^^^^^^
 
+- ``photutils.isophote``
+
+  - Fix a bug that ``build_ellipse_model`` falls into an infinite loop
+    when the pixel to fit is outside of the image. [#1038]
+
 - ``photutils.segmentation``
 
   - Fixed a bug where ``source_properties`` would fail with unitless
--- a/photutils/isophote/model.py
+++ b/photutils/isophote/model.py
@@ -128,7 +128,6 @@ def build_ellipse_model(shape, isolist, fill=0., high_harmonics=False):
             i = int(x)
             j = int(y)
 
-            # if outside image boundaries, ignore.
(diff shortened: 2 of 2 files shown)


larrybradley:
@ibusko Can you please review?

larrybradley:
Thanks, @ysBach

CHANGES.rst
```diff
@@ -17,6 +17,11 @@ New Features
 Bug Fixes
 ^^^^^^^^^
 
+- ``photutils.isophote``
+
+  - Fix a bug that ``build_ellipse_model`` falls into an infinite loop
+    when the pixel to fit is outside of the image. [#1038]
```
larrybradley, line 23:
```suggestion
    when the pixel to fit is outside of the image. [#1039]
```

The changelog should contain the PR number where the bug was fixed, not the issue number.

larrybradley:
The travis-ci test failure is unrelated to this PR.  I'll fix that in a separate PR.

PR #1060, opened 2020-07-22
Remove unnecessary check for warning class

larrybradley:
The check for the warning category is unnecessary since only that warning type is caught in the context manager.  Somehow this test started failing in travis-ci, but I cannot reproduce the failure.

### DIFF
--- a/photutils/detection/tests/test_core.py
+++ b/photutils/detection/tests/test_core.py
@@ -5,7 +5,8 @@ Tests for the core module.
 
 import warnings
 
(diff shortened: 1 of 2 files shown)


larrybradley:
The root cause of the CI failures is the "Distutils was imported before Setuptools" warning raised in `setuptools 49.2.0` (xref https://github.com/astropy/astropy/pull/10576).  `photutils` does not import `distutils` so the error is coming from dependent packages.  There does not appear to be a way to suppress this error (I've tried), so I've temporarily removed the tests with `catch_warnings` that were failing due to this issue.

Issue #707, opened 2018-08-17
Input Issue Resulting in Infinite Loop


HSouch:
On extremely rare occasions, (3 in ~16000 iterations for my own work), the fitting routine in photutils.isophote.fitter results in an infinite loop. My own code that uses the fitting routine is below.
```
flux = Ellipse(data)  # Where data is a valid numpy ndarray with no nan or inf pixels

try:
    fitting_list = flux.fit_image()
except IndexError:
    fitting_list = []
...
```

When this infinite loop occurs, it continuously logs the following:
```
INFO: Improper input: N=5 must not exceed M=0 [photutils.isophote.fitter]
INFO: Improper input: N=5 must not exceed M=0 [photutils.isophote.fitter]
...
```
which appears to be a result of the code on lines 151-155. Perhaps changing the way this exception is handled to a warning would be a better alternative?

larrybradley:
CC: @ibusko

shivangeerathi:
Hi, 
I am having the same issue with the latest version of photutils (v0.7.2) on a 64-bit machine with Ubuntu 18.04 LTS. It occurs rather randomly. 
Does anyone knows a fix to this ?

ibusko:
@HSouch @shivangeerathi  could you please provide us with a sample image were the issue manifests itself?

shivangeerathi:
@ibusko 
[Here](https://we.tl/t-b1Aicw29X9) is one of the cases where the error occurs.

ibusko:
@shivangeerathi I can't make it go into an infinite loop. Instead, it fails at the very beginning with a stop code of -1. Running under a debugger, I can see that this is caused by a positive intensity radial gradient condition. This gradient must be always negative for the algorithm to work (intensity decreases as the radius increases). The cause in this particular image seems to be one (or two) star images that are superimposed on the galaxy image. The mask file you sent me doesn't seem to mask out these two particular star images.

That said, you probably have found a combination of input parameters that makes the algorithm successfully go over this specific hurdle. And then it will go into the loop. What is the specific set of parameter values you use?

shivangeerathi:
This is the input geometry that I had used: 
`x0 = 1000, y0 = 1000,  sma = 50,  eps = 0.3,  pa = 135*np.pi/180`
In addition, I had initially restricted the maxsma to 670 px.

ibusko:
[...]

PR #1091, opened 2020-10-10
Changed "info" log to warning to allow for infinte loop catching.

HSouch:
In the below issue thread, I mentioned an infinite loop that can rarely occur when using photutils for elliptical isophote analysis.
https://github.com/astropy/photutils/issues/707

Changing the log message from info to a warning allows users to catch it using ```warnings.filterwarnings("error")```.

### DIFF
--- a/photutils/isophote/fitter.py
+++ b/photutils/isophote/fitter.py
@@ -160,7 +160,7 @@ class EllipseFitter:
                 coeffs = fit_first_and_second_harmonics(values[0], values[2])
                 coeffs = coeffs[0]
             except Exception as e:
-                log.info(e)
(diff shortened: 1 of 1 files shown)


HSouch:
An example cutout that causes this issue (at least on my end) with a current version of ```photutils``` is included below.
[bad_cutout.zip](https://github.com/astropy/photutils/files/5358310/bad_cutout.zip)

larrybradley:
cc: @ibusko

ibusko (approved):
Looks like a good solution. Thanks for the fix!

larrybradley (approved):
Thanks, @HSouch.

PR #1093, opened 2020-10-12
Remove aperture benchmarks

larrybradley:
I ran `bandit` (security analyzer) on the code and it reports 3 issues, all in the `bench_aperture.py` module (all related to its use of `pickle`).  I'm going to remove this file.  We haven't actually been running benchmarks (for quite some time), and in fact the benchmarks do not currently even run due to API changes.  If we want to do benchmarking in the future, we should probably use `airspeed velocity` like https://github.com/astropy/astropy-benchmarks.


bsipocz (approved):
Totally agree with all what you have said. As I recall these dates back to when `asv` wasn't a thing yet, and mostly used for my gsoc, but not beyond that.

It's also somewhat sad that you needed to run a separate sec. analyser to pick up the issues, I have assumed for a while now that github does it now by default (yet the issues were not picked up).  I'll give `bandit` a try then for my repos.

larrybradley:
The CI failures are unrelated.  Something in the CI broke since Friday.  @pllim thinks it's related to the astropy 4.0.2 release.

pllim:
Maybe you should enable RTD PR build, then you don't need Travis CI to pass for the docs.

pllim:
Oh, ops... I thought this was the other PR. LoL. Sorry for the noise.

Issue #468, opened 2016-11-23
Doesn't build


jankansky:
If I run:

git clone https://github.com/astropy/photutils.git
cd photutils
sudo python ./setup.py install

The build fails at:

running build_ext
error: [Errno 2] Could not find C/C++ file photutils/geometry/elliptical_overlap.(c/cpp) for Cython file photutils/geometry/elliptical_overlap.pyx when building extension photutils.geometry.elliptical_overlap. Cython must be installed to build from a git checkout.: 'photutils/geometry/elliptical_overlap.c'

mirca:
Could you please provide your astropy, photutils, and numpy versions?

larrybradley:
@jankansky  To build `Photutils` from the source, you must have `Cython` installed.  According to the warning message, you don't have `Cython` installed.  See the prerequistes here: http://photutils.readthedocs.io/en/latest/photutils/install.html#prerequisites

jankansky:
Thank you.  Cython was not installed.

PR #1099, opened 2020-10-15
add friendly messages about missing Cython or extension_helpers python packages

dstndstn:
just try to import them in setup.py immediately before use, and print a helpful error message.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,17 +76,14 @@ except Exception:
 import os  # noqa
 from setuptools import setup  # noqa
 
-try:
-    import Cython
-except ImportError:
-    print('To build photutils, you need the Cython package.  Try "pip install Cython" or similar.')
-    sys.exit(-1)
-
-try:
-    import extension_helpers
-except ImportError:
-    print('To build photutils, you need the extension_helpers package.  Try "pip install extension_helpers" or similar.')
-    sys.exit(-1)
+import importlib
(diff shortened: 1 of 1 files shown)


larrybradley:
@dstndstn Why is this needed?  We use PEP518 (`pyproject.toml`) to define build-time dependencies.  `pip install` should pick up any needed build-time dependencies.

larrybradley:
Also see https://github.com/astropy/astropy-APEs/blob/master/APE17.rst

dstndstn:
I ran into this because I checked out the git repo and did a "python3
setup.py install"; what happened first was
https://github.com/astropy/photutils/issues/468
(which is quite a bad failure mode -- for me it didn't print anything
suggesting that cython was expected but missing)
and then the missing extension_helpers.

larrybradley:
`python setup.py install` is generally no longer recommended to install python packages.  To install from source, use `pip install .[all]`.  I will update the install docs for dev installs.

dstndstn:
news to me!

pllim:
This isn't a new problem. See astropy/astropy#9965 for discussion. It is better to use new command than to overcomplicate `setup.py`.

bsipocz:
> news to me!

There has been some significant improvements for python packaging and pip in the recent years, this is one of the nice outcomes, no more confusions about setup.py and whether to use build/install/develop, pip install is now always the answer.

PR #1103, opened 2020-10-22
modify _data_coords calculation

jdermigny:
Using numpy.indices is a factor of 25 faster than the itertools.product method. This is a dramatic improvement for the Background2D execution time when analyzing large (2048x2048) images.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,15 @@ New Features
 1.0.2 (unreleased)
 ------------------
 
+New Features
+^^^^^^^^^^^^
+
+- ``photutils.background``
+
+  - Improved the performance of ``Background2D`` (e.g., by a factor
+    of ~4 with 2048x2048 input arrays when using the default interpolator).
+    [#1103]
+
 Bug Fixes
 ^^^^^^^^^
 
--- a/photutils/background/background_2d.py
+++ b/photutils/background/background_2d.py
@@ -746,7 +746,7 @@ class Background2D:
 
         # the position coordinates used when calling an interpolator
         ny, nx = self.data.shape
-        self.data_coords = np.array(list(product(range(ny), range(nx))))
+        self.data_coords = np.indices((nx, ny)).T.reshape(nx*ny, 2)[:, [1, 0]]
 
     @lazyproperty
     def mesh_nmasked(self):


jdermigny:
@bsipocz @larrybradley is there anything else you would like me to do for this pull request?

larrybradley:
Thanks, @jdermigny.  Can you please add an entry in `CHANGES.rst` under `1.1.0` "New Features" in a new ``photutils.background`` section?  It should note the improved performance of `Background2D` and include the PR number in brackets at the end.  I will review the PR tomorrow.

jdermigny:
The build failure I believe is related to the recent BkgIDWInterpolator changes. Let me know if you need me to rebase.

CHANGES.rst
```diff
@@ -9,6 +9,9 @@ New Features
   - The ``PixelAperture.plot()`` method now returns a list of
     ``matplotlib.patches.Patch`` objects. [#923]
 
+- ``photutils.background``
+  - Improved the performance of ``Background2D._calc_coordinates`` by a factor of ~ 25. [#1103]
```
larrybradley, line 13:
```suggestion
- ``photutils.background``

  - Improved the performance of ``Background2D`` (e.g., by a factor
    of ~4 with 2048x2048 input arrays when using the default interpolator).
    [#1103]
```

Much better to report the overall performance improvement instead of that on a private method.

photutils/background/background_2d.py
```diff
@@ -745,8 +745,8 @@ def _calc_coordinates(self):
         self.yx = np.column_stack([self.y, self.x])
 
         # the position coordinates used when calling an interpolator
-        ny, nx = self.data.shape
-        self.data_coords = np.array(list(product(range(ny), range(nx))))
+        nx, ny = self.data.shape
```
larrybradley, line 748: This was fixed in #1104.  You'll need to rebase to pick up those changes.  The correct order is ``ny, nx``.

photutils/background/background_2d.py
```diff
@@ -745,8 +745,8 @@ def _calc_coordinates(self):
...
+        self.data_coords = np.indices((ny, nx)).T.reshape(nx*ny, 2)
```
larrybradley, line 749:
```suggestion
        self.data_coords = np.indices((nx, ny)).T.reshape(nx*ny, 2)[:, [1, 0]]
```

While reviewing this PR, I discovered that both this change and the version in master were both incorrect because of the order of the coordinates.  I fixed the version in master in #1104 and added a simple regression test.  That's the currently failing test, and it's a real test failure.  The suggested changes above fix this issue (after you rebase on the current master).

The simple test is:
```python
from photutils import Background2D, BkgIDWInterpolator
data = np.arange(12).reshape(3, 4)
interpolator = BkgIDWInterpolator()
bkg = Background2D(data, (1, 1), filter_size=1,
                   interpolator=interpolator)
bkg.background
```
The result should the same as the input `data`.

larrybradley (changes requested):
Thanks, @jdermigny.  See the inline comments for the requested changes.

larrybradley (approved):
Thanks, @jdermigny.

PR #1104, opened 2020-10-27
Fix bug in Background2D when using BkgIDWInterpolator

larrybradley:
This PR fixes a bug I discovered (while reviewing #1103) where `Background2D` with `BkgIDWInterpolator` would give incorrect results for the full-sized interpolated background image.


larrybradley:
The test failure is unrelated.

PR #1121, opened 2020-11-17
DOC: Update license year 

pllim:
Update license year.

But more importantly, to see if `[ci skip]` really works on Actions.


pllim:
@larrybradley , not sure when it last worked for you but it doesn't anymore.

larrybradley (approved):
Too bad it `skip-ci` doesn't work anymore.

I had removed the end license year because I think it's unnecessary.

Issue #1128, opened 2020-12-02
ellipse.fit_image() gets stuck: "INFO: Improper input: N=5 must not exceed M=0 [photutils.isophote.fitter]"


davidjsetton:
Hi,

I am using photutils to fit isophotal ellipses to galaxy images from the hyper-suprime cam survey. However, occasionally, I run into an issue where the fit_image() module prints a series of lines similar to:

"INFO: Improper input: N=5 must not exceed M=0 [photutils.isophote.fitter]"

This printing goes on forever until I manually stop the code from running. Sometimes, I can catch this because it spits out an IndexError, but other times, it occurs without the error.

I've recreated the error in a notebook and have attached all the relevant files. Please let me know if there is anything else I can do to help you locate the error.

Best,

David

[photutils_errors.zip](https://github.com/astropy/photutils/files/5631617/photutils_errors.zip)

larrybradley:
@davidjsetton Thanks for reporting.  @ibusko -- can you take a look?

ibusko:
@davidjsetton thanks for the detailed error report! It's always helpful when the report comes with a reproducible setup.

I found that the infinite loop condition is triggered when the sample extraction function in the isophote code returns a zero-length sample. In your specific case, this happens when the ellipse being sampled has very small radius If you set `minsma=2.5` in the `deep_ellipse.fit_image` command, it will stop before reaching the point where the sample is zeroed (at `sma=2.3325`). 

I will put in a fix for this, for the next release. In the meantime, I hope the `minsma` solution works for you. If not, please let me know so we can think about a different approach.

davidjsetton:
@ibusko thanks so much for the quick reply! 

That definitely makes sense. I was actually able to suppress the behavior by forcing the initial guess for the fitter to the actual brightest pixel rather than the pixel which "should" be the brightest pixel; it must have been starting on a solution with a bad enough center that it ran into some issues with the inner isophotes. Making this change has increased my rate of convergence by a lot.

However, it's still an issue for some of the other images in the dataset. For example, I've attached another notebook which is identical to the previous one but with another object in the sample, and this one fails even with the minsma=2.5 solution. I'm totally fine with skipping these objects in the fits for now, but I think I'm going to need a solution which can catch them when they occur, because otherwise I get stuck in the infinite loop.

[photutils_errors.zip](https://github.com/astropy/photutils/files/5638022/photutils_errors.zip)

[...]

PR #1129, opened 2020-12-03
Fix bug with zero length samples 

ibusko:
Bug was reported by a user (issue #1128)


larrybradley (approved):
Thanks, @ibusko.  Since this is bug fix, I'll put it in v1.0.2.  I'll move the changelog entry.  Thanks!

PR #1135, opened 2020-12-21
formerly hacky fix for incorrect model names trying to be accessed if more than two stars in group

krachyon:
After running into https://github.com/astropy/photutils/issues/845 I tried to fix it generically (https://github.com/krachyon/photutils/tree/parameter_name_matching) by comparing the parameter names of the psf model and matching them to those that should be output in `_model_params2table()` but ran into the issue that sometimes a compound model seems to have the parameters combined by adding a postfix (param_i->param_i_j) or by incrementing a existing postfix(param_i -> param_j). More Details on that train of thought in the comment in `photometry.py`

I decided it's not really worth trying to predict that behaviour and came up with this pretty hacky way to handle both. I'm not sure if this is something merge worthy, but for now it seems to solve the issue on my end. 

I guess the proper solution would involve modifying the way that the expected parameter names are derived in the combination of `prepare_psf_model` and `photometry._define_fit_param_names()` `get_grouped_psf_model()`. I suspect in the latter function we could always compare the before-after state, but ideally there should probably be a way to access the parameters of submodels by their original names in a compound model implemented in astropy.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -81,6 +81,9 @@ Bug Fixes
   - Fixed a bug in ``EPSFBuild`` where a warning was raised if the input
     ``smoothing_kernel`` was an ``numpy.ndarray``. [#1146]
 
+  - Fixed a bug that caused photometry to fail on an ``EPSFmodel`` with
+    multiple stars in a group. [#1135]
+
 - ``photutils.segmentation``
 
   - Fixed ``SourceProperties`` ``local_background`` to work with
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -505,7 +505,7 @@ class BasicPSFPhotometry:
             in this package like `~photutils.psf.sandbox.DiscretePRF`,
             `~photutils.psf.IntegratedGaussianPRF`, or any other
             suitable 2D model.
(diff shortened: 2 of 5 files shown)


krachyon:
So according to tox -e codestyle line length is limited to 100, but the pep8 bot wants 79. What's the correct amount?

larrybradley:
Thanks, @krachyon.  Since this has been a long-term issue, I'm inclined to merge it and think about if there's a better solution later.  @eteq - would you like to review this?

krachyon:
Ok, so I managed to trigger this issue again after 'fixing' it and looking at it again I found a solution that's not actually that awful. Instead of trying to soothsay the parameter name, I tagged the models belonging to individual stars during combination so that later we just have to look for the name.

larrybradley (changes requested):
Thanks, @krachyon.  I'm not an expert in this part of the code, but since this fixes the issue I'm inclined to merge it.  I left a couple style comments.

Can you also please add a changelog entry in the v1.1.0 Bug Fixes section under `photutils.psf`?

larrybradley (approved):
Thanks again, @krachyon.

PR #1144, opened 2021-01-14
Change Default EPSFBuilder Recentering Function

Onoddil:
As per several issues raised, but most recently #969, there remains an outstanding issue in some cases with ePSF building. It seems that the issue now lies in the recentering function used during the process, which was left as `centroid_epsf` after #974 fixed some issues with the original bug fix of #817.

Thus this PR is a minor one to change the default `recentering_func` within `EPSFBuilder` to `centroid_com`, reverting the implementation of `centroid_epsf`. This seems to solve the issue raised in #969, and likely closes several related issues at the same time.

The change makes our unit tests converge a little bit slower than they did with `centroid_epsf`, so I tweaked a few parameters in the various `EPSFBuilder` calls to pass again. I also wasn't sure whether this was an API change or a bug fix, so let me know if this should really be an API change rather than a bug fix in the changelog.

cc @larrybradley as follow on from the discussion on #969

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -74,7 +74,7 @@ Bug Fixes
 - ``photutils.psf``
 
   - Changed the default ``recentering_func`` in ``EPSFBuilder``, to avoid
-    convergence issues. [#1143]
+    convergence issues. [#1144]
 
 
 1.0.1 (2020-09-24)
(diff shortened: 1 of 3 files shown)


larrybradley (approved):
Thanks, @Onoddil!  This looks good to me.  Changing a parameter default is typically not a minor change, but since this fixes several issues I'm inclined to call it a bug fix and put it in 1.0.2 (as you have it).  Thanks!

PR #1146, opened 2021-01-15
don't cause warning when using custom smoothing kernel in EPSFBuilder

krachyon:
Minor improvement to avoid warning message by reordering comparisons


larrybradley (approved):
Thanks, @krachyon!

larrybradley:
Changelog entry added here: dd193883ad0c2bf4f8ca993ab4ea7e4c9ba93add

PR #1152, opened 2021-01-28
Make EPSFStars usable with multiprocessing

krachyon:
I've encountered an issue where I couldn't return EPSFStars through multiprocessing, due to the custom `__getattr__` causing infinite recursion with `multiprocessing.ForkingPickler`. It does not seem to be an issue with the regular pickle module due to some difference of how it tries to find `__getstate__` which I don't really understand...

The included test is probably not something worthwhile to keep, but should reproduce the problem.

BTW are these tiny PRs fine or should I wait until/if I get more "content"?

### DIFF
--- a/photutils/psf/epsf_stars.py
+++ b/photutils/psf/epsf_stars.py
@@ -321,6 +321,13 @@ class EPSFStars:
         for i in self._data:
             yield i
 
+    # explicit set/getstate to avoid infinite recursion from pickler using __getattr__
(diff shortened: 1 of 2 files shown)


larrybradley (approved):
Thanks, @krachyon.  This needs a changelog entry, but I can add that.

krachyon:
Oh, sorry, wasn't aware that the changelog was something that should be done with the pull-request itself. Will add it in the future

PR #1178, opened 2021-03-09
MNT: Replace master with main

pllim:
This PR attempts to replace all mention of `master` to `main` in this repository. This should be reviewed and merged right after the maintainer has renamed the default branch.

This is an automated update made by the `batchpr` tool :robot: - feel free to close if it doesn't look good! You can report issues to @pllim.

This is part of #1176


larrybradley:
An additional note that this PR also requires renaming the default branch in the `photutils-datasets` repo.

larrybradley (approved):
Thanks, @pllim!

PR #1204, opened 2021-04-13
Update RTD config

larrybradley:
RTD is now configured to build on PRs.  This PR updates the RTD config to fail on warnings  and updates python to ~3.9~ 3.8.

EDIT:  This PR also removes `build_docs` from the GitHub Actions CI tests.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -3,8 +3,13 @@ version: 2
 build:
   image: latest
 
+sphinx:
+  builder: html
+  configuration: docs/conf.py
+  fail_on_warning: true
+
 python:
-  version: 3.7
+  version: 3.8
   install:
     - method: pip
       path: .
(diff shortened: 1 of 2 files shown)


.readthedocs.yml
```diff
@@ -3,8 +3,13 @@ version: 2
...
   image: latest
 
+sphinx:
+  builder: html
+  configuration: docs/conf.py
+  fail_on_warning: true
+
 python:
-  version: 3.7
+  version: 3.9
```
pllim, line 12: I remember trying Python 3.9 a while back but it failed for some reason and I had to fall back to 3.8. If it works now, great!

larrybradley: 🤞

larrybradley: Falling back to 3.8

PR #1207, opened 2021-04-20
Remove deprecated subpixel keyword in find_peaks

larrybradley:
The deprecated `subpixel` keyword was removed two years ago in https://github.com/astropy/photutils/pull/835. While the functionality of the keyword was removed along with its documentation, unfortunately the keyword was not removed from the function signature.  This PR fixes that issue.


larrybradley:
The devdeps CI test failure is a known upstream issue with astropy sigma_clip.

PR #1231, opened 2021-08-26
Add UserStarFinder class

stsci-hack:
This implements a new class/module called UserStarFinder which allows users to generate a catalog based on source positions they provide from a separate external source using the DAOStarFinder code to perform the photometry. This enables users to employ any method they want to identify the sources, including ePSFs or Laplacian-of-Gaussian functions or model PSFs.  A copy of this code is being used in Drizzlepac to generate the catalogs for the Hubble Advanced Products (HAP) single-visit-mosaic (SVM) images.

### DIFF
--- a/photutils/detection/__init__.py
+++ b/photutils/detection/__init__.py
@@ -7,5 +7,6 @@ image.
 from .base import *  # noqa
 from .daofinder import *  # noqa
 from .irafstarfinder import *  # noqa
+from .userstarfinder import *  # noqa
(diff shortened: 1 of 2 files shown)


larrybradley:
Thanks, @stsci-hack.  Since `UserStarFinder` is essentially the same thing as `DAOStarFinder` but with the added `coord` keyword, I think a better solution is simply add the `coord` keyword to `DAOStarFinder`. Would that work for you?

PR #1251, opened 2021-09-21
Make result of psf-photometry re-usable as initial guess

krachyon:
when doing something like

```python
  result_table = photometry(image)
  result_table['x'] = result_table['x_fit']
  result_table['y'] = result_table['y_fit']
  result_table['flux'] = result_table['flux_fit']
  second_result = other_photometry(image, result_table)
```

a bunch of errors occur due to the re-naming of columns by `Table.hstack`, trying to create existing column `group_id` and caring about the order of the `id` column. These changes should enable the `Photometry`-classes to update tables with the fields already defined.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,14 @@ New Features
   - Added a ``xycoords`` keyword to ``DAOStarFinder`` and
     ``IRAFStarFinder``. [#1248]
 
+- ``photutils.psf``
+
+  - Enabled the reuse of an output table from ``BasicPSFPhotometry`` and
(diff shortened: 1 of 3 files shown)


larrybradley (approved):
Thanks, @krachyon.  I went ahead and updated this PR directly so that it can go in the 1.2.0 release.  I'm calling this a feature enhancement instead of bug because it's clear the original author of the PSF-fitting code did not intend for the user to input a table with a `group_id` column.  With this PR, an `init_guesses` table with a `group_id` column will allow the `group_maker` step to be skipped.

PR #1267, opened 2021-11-10
To fix the documentation of deblend_sources function

### LINKED ISSUE #1266: deblend_sources function documentation
This is likely a minor bug, 1.2 changed `filter_kernel` to `kernel`. But the `deblend_sources` function description says `filter`. I think this should be changed to `kernel`, similar to the `detect_sources` function. 

    filter : array-like or Kernel2D, optional

    The array of the kernel used to filter the image before thresholding. Filtering the image will smooth the noise and maximize detectability of objects with a shape similar to the kernel.

https://photutils.readthedocs.io/en/stable/api/photutils.segmentation.deblend_sources.html#photutils.segmentation.deblend_sources

prajwel:
This pull request addresses issue #1266. 

Closes #1266.

### DIFF
--- a/photutils/segmentation/deblend.py
+++ b/photutils/segmentation/deblend.py
@@ -47,7 +47,7 @@ def deblend_sources(data, segment_img, npixels, kernel=None, labels=None,
         that an object must have to be detected.  ``npixels`` must be a
         positive integer.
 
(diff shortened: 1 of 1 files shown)


larrybradley (approved):
Thanks, @prajwel.

PR #1281, opened 2021-12-31
MNT: Stop using deprecated astropy.tests stuff

pllim:
This pull request is to remove soon to be deprecated code; also see astropy/astropy#12633 .

* MNT: Stop using deprecated astropy.tests stuff
* TST: Turn unhandled warnings into exceptions.

### DIFF
--- a/photutils/centroids/tests/test_core.py
+++ b/photutils/centroids/tests/test_core.py
@@ -4,7 +4,7 @@ Tests for the core module.
 """
 
 import itertools
-import warnings
+from contextlib import nullcontext
 
 from astropy.modeling.models import Gaussian2D
 from astropy.utils.exceptions import AstropyUserWarning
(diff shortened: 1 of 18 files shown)


photutils/centroids/tests/test_core.py
```diff
@@ -82,24 +82,26 @@ def test_centroid_com_nan_withmask(use_mask):
...
-        assert yc > yc_ref
-    assert len(warnlist) == nwarn
+    assert_allclose(xc, xc_ref, rtol=0, atol=1.e-3)
+    assert yc > yc_ref
     if nwarn == 1:
-        assert issubclass(warnlist[0].category, AstropyUserWarning)
+        assert len(warnlist) == nwarn
 
-    with warnings.catch_warnings(record=True) as warnlist:
+    with pytest.warns(AstropyUserWarning,
```
pllim, line 99: This warning always is emitted now that we don't use `warnings.catch_warnings`. You might want to double check if this is the intended behavior of photutils or a bug.

photutils/detection/tests/test_peakfinder.py
```diff
@@ -68,12 +65,10 @@ def test_npeaks(self):
 
     def test_border_width(self):
         """Test border exclusion."""
-        with catch_warnings(NoDetectionsWarning) as warning_lines:
+        with pytest.warns(NoDetectionsWarning,
+                          match='No local peaks were found'):
             tbl = find_peaks(PEAKDATA, 0.1, box_size=3, border_width=3)
             assert tbl is None
-            assert len(warning_lines) > 0
```
pllim, line 74: No need for this anymore as `pytest.warns` will throw error if the warning is not there.

photutils/detection/tests/test_peakfinder.py
```diff
@@ -118,28 +113,25 @@ def test_wcs_values(self):
...
         """
         Test for an empty output table with the expected column names
         when no peaks are found.
         """
-        with catch_warnings(NoDetectionsWarning):
+        with pytest.warns(NoDetectionsWarning):
             tbl = find_peaks(IMAGE, 10000)
             assert tbl is None
-
+        with pytest.warns(NoDetectionsWarning):
```
pllim, line 128: I think checking for the warning in every call that it is expected from is more correct, hence the extra lines. Otherwise, you will not know if one of them suddenly stops emitting the warning if they are all dumped into the same context manager.

photutils/detection/tests/test_peakfinder.py
```diff
@@ -148,7 +140,4 @@ def test_data_nans(self):
         """Test that data with NaNs does not issue Runtime warning."""
         data = np.copy(PEAKDATA)
         data[1, 1] = np.nan
-
-        with warnings.catch_warnings():
```
pllim, line 152: This is no longer needed because we are turning all uncaught warnings into exceptions in `setup.cfg`.

photutils/isophote/tests/test_model.py
```diff
@@ -62,20 +62,20 @@ def test_model_minimum_radius():
...
-                                      step=0.4, nclip=3)
-
-    model = build_ellipse_model(data.shape, isophote_list,
-                                fill=np.mean(data[0:50, 0:50]))
-
-    # It's enough that the algorithm reached this point. The
-    # actual accuracy of the modelling is being tested elsewhere.
-    assert data.shape == model.shape
+    filepath = get_pkg_data_filename('data/minimum_radius_test.fits')
+    with fits.open(filepath) as hdu:
```
pllim, line 66: This is to get rid of the unclosed file warning.

photutils/psf/tests/test_photometry.py
```diff
@@ -152,6 +151,9 @@ def test_psf_photometry_niters(sigma_psf, sources):
         assert_array_equal(cp_image, image)
 
 
+@pytest.mark.filterwarnings('ignore:Both init_guesses and finder are '
```
pllim, line 154: Blanket ignore here because these warnings are not consistent for all the cases that pass through `parametrize`.

larrybradley (approved):
Many thanks, @pllim!

PR #1298, opened 2022-02-18
Add curve-of-growth-based FWHM estimator

eteq:
## TODO

- [ ] Name a Jdaviz dev to take over this PR.
- [ ] Compare this to what is already in imexam. Which one should we use?
- [ ] Finalize API.
- [ ] Implement chosen algorithm.
- [ ] Add, update, and fix tests.
- [ ] Add and update documentation.
- [ ] Update change log, if necessary.

## Original post

This adds a function that basically implements the algorithm I suggested for a project that wants to *use* photutils: spacetelescope/jdaviz#1048 - I described the algorithm there, but it's basically "put down an aperture, consider "max" to be all of the flux, then find the radius that provides half of that flux to be the half-width-at-half-max.

This isn't the most robust algorithm in the world for FWHMs, but it is probably conceptually the simplest.

This is a working PR, but I should probably document it a bit better and do more comprehensive tests, so I'm opening it for now as a draft.  @larrybradley in particular might want to comment if you think I've put this in the right sub-package... 

cc @pllim @larrybradley

### DIFF
--- a/photutils/aperture/photometry.py
+++ b/photutils/aperture/photometry.py
@@ -142,7 +142,7 @@ def aperture_photometry(data, apertures, error=None, mask=None,
         nddata_attr = {'error': error, 'mask': mask, 'wcs': wcs}
         for key, value in nddata_attr.items():
             if value is not None:
-                warnings.warn(f'The {key!r} keyword is be ignored. Its value '
+                warnings.warn(f'The {key!r} keyword will be ignored. Its value '
                               'is obtained from the input NDData object.',
                               AstropyUserWarning)
 
--- a/photutils/morphology/fwhm.py
+++ b/photutils/morphology/fwhm.py
@@ -1,0 +1,184 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+This module defines tools to perform aperture photometry.
+"""
+import warnings
+
+import numpy as np
+from astropy.nddata import NDData, StdDevUncertainty
+import astropy.units as u
+from astropy.utils.exceptions import AstropyUserWarning
+
+from photutils.aperture import Aperture, CircularAperture, SkyCircularAperture, SkyAperture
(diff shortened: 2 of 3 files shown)


eteq:
hah pep8speaks is very unhappy with me.  Good thing I started with a draft 🤖 😠

pllim:
We need to compare this with what is already established in `imexam`, see https://github.com/spacetelescope/imexam/blob/b3975733e0ef459fc7d137acec9bf4aa2865d6c5/imexam/imexamine.py#L1266 . It also uses `photutils` for photometry.

pllim:
I am guessing this is going to be a blocker for both spacetelescope/jdaviz#1048 and spacetelescope/jdaviz#1049 ?

Can we even use this without a way to translate between `regions` region and `photutils` region first?

pllim:
@eteq , we discussed this at Indigo tag-up today. I modified your original post with our plan. Please look at it and let us know if you have questions or concerns.

eteq:
> We need to compare this with what is already established in imexam,

Honestly I am *not* convinced we should give the imexam implementation precedence, as the approach in this PR in some ways is intentionally different (e.g., how it treats fractional pixels).  I agree a comparison *does* make sense for sanity-checking and to ensure I'm not making a mistake in something here, but the answer might still be "use this instead of imexam" even if they are different.

Rest of the plan looks good to me!  Although even if someone else can/wants to take this over I still care a lot about the answer to this algorithm ion `photutils` regardless of how it gets used in `jdaviz`.

eteq:
Oh, and I also think this is already in the right direction - the implementation and API are basically complete modulo any concrete changes.  But I would be very happy to get some help making more comprensive tests, as then I could focus just on docs!

larrybradley:
I'm going to go ahead and close this since there are now several ways to do this in photutils.

Issue #1316, opened 2022-03-22
ApertureStats not optimized for large array even when aperture is small


pllim:
# Example use case

Download `hlsp_candels_hst_wfc3_gs-tot_f105w_v1.0_drz.fits` (4.9 GB) from https://archive.stsci.edu/missions/hlsp/candels/goods-s/gs-tot/v1.0/

```python
from astropy.io import fits
from astropy.wcs import WCS

from photutils.aperture import ApertureStats, CircularAperture

filename = 'hlsp_candels_hst_wfc3_gs-tot_f105w_v1.0_drz.fits'
pf = fits.open(filename)

data = pf[0].data  # (40500, 32400)
w = WCS(pf[0].header)

# This was interactively drawn around a real source
aperture = CircularAperture((12339.6, 16491.9), 33.6)  # area = 3546.7

# This takes a long time (~30 sec) and ate up a few GB memory
# that was not freed after calculations are done.
apstats = ApertureStats(data, aperture, wcs=w, local_bkg=0)

pf.close()
```

larrybradley:
Thanks, working on a fix...

larrybradley:
I could not reproduce your 30-sec runtime.  On `main` it ran for me in 7.37 sec.  With the fix in https://github.com/astropy/photutils/pull/1318 it now runs in 220 ms.

PR #1338, opened 2022-05-04
DOC: Encourage local_bkg in ApertureStats

pllim:
No reason not to use ``local_bkg``, right? Should put best practice in example? xref:

* #1316

### DIFF
--- a/docs/aperture.rst
+++ b/docs/aperture.rst
@@ -375,6 +375,13 @@ subtract the background from the data::
 
     >>> phot_table = aperture_photometry(data - bkg, aperture)  # doctest: +SKIP
 
+In the case of a constant background, you can pass it in as global background
+using ``local_bkg`` in :class:`~photutils.aperture.ApertureStats`, which
+would prevent the need of reading the entire memory-mapped data into memory
+before hand, as what would happen if you subtract it manually above;
+so instead you could do this::
+
+    >>> aperstats = ApertureStats(data, aperture, local_bkg=bkg)  # doctest: +SKIP
 
 Local Background Subtraction
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/aperture.rst
+++ b/docs/aperture.rst
@@ -375,11 +375,11 @@ subtract the background from the data::
 
     >>> phot_table = aperture_photometry(data - bkg, aperture)  # doctest: +SKIP
 
-In the case of a constant background, you can pass it in as global background
-using ``local_bkg`` in :class:`~photutils.aperture.ApertureStats`, which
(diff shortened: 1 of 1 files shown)


pllim:
Hmm, `local_bkg` throws exception when aperture has multiple centers. Is this feature or bug? From the doc example:

```python
>>> aper2
<CircularAperture([[150.,  25.],
                   [ 90.,  60.]], r=10.0)>
>>>  median
5.408250014261292
>>> aperstats2 = ApertureStats(data, aper2, local_bkg=median)
```
```
.../photutils/aperture/stats.py in __init__(self, data, aperture, error, mask, wcs, sigma_clip, sum_method, subpixels, local_bkg)
    228                 raise ValueError('local_bkg must be a 1D array')
    229             if len(local_bkg) != self.n_apertures:
--> 230                 raise ValueError('local_bkg must have the same length as '
    231                                  'the aperture')
    232             if np.any(~np.isfinite(local_bkg)):

ValueError: local_bkg must have the same length as the aperture
```

larrybradley:
Definitely a feature.  It's a **local** background (i.e. around a particular position), not a global background.

larrybradley:
From the docstring:
" The local_bkg values correspond to the input aperture positions. local_bkg must have the same as the length of the input aperture."

pllim:
Thanks for the hint. I fixed the doctest but I'll leave it up to you if the changes make sense.

pllim:
@larrybradley , I updated the doc as you requested but not sure if I made it sound too confusing. What do you think?

docs/aperture.rst
```diff
@@ -375,6 +375,13 @@ subtract the background from the data::
 
     >>> phot_table = aperture_photometry(data - bkg, aperture)  # doctest: +SKIP
 
+In the case of a constant background, you can pass it in as global background
+using ``local_bkg`` in :class:`~photutils.aperture.ApertureStats`, which
+would prevent the need of reading the entire memory-mapped data into memory
+before hand, as what would happen if you subtract it manually above;
+so instead you could do this::
```
larrybradley, line 382:
```suggestion
In the case of a constant global background, you can pass in the background
value using ``local_bkg`` in :class:`~photutils.aperture.ApertureStats`.
This would avoid reading an entire memory-mapped array into memory
beforehand, as would happen if you manually subtract the background as
shown above. So instead you could do this::
```

larrybradley (commented):
How about this?

docs/aperture.rst
pllim: Thanks!

larrybradley (approved):
Thanks, @pllim!

PR #1426, opened 2022-10-21
Progressbar in PSF photometry

keflavich:
This PR adds an option to use a progressbar when doing PSF photometry.

### DIFF
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -243,12 +243,12 @@ class BasicPSFPhotometry:
                               'automatically ignored.', AstropyUserWarning)
         return mask
 
-    def __call__(self, image, *, mask=None, init_guesses=None):
+    def __call__(self, image, *, mask=None, init_guesses=None, **kwargs):
         """
         Perform PSF photometry. See `do_photometry` for more details
         including the `__call__` signature.
         """
-        return self.do_photometry(image, mask=mask, init_guesses=init_guesses)
+        return self.do_photometry(image, mask=mask, init_guesses=init_guesses, **kwargs)
 
     def do_photometry(self, image, *, mask=None, init_guesses=None, **kwargs):
         """
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -401,7 +401,7 @@ class BasicPSFPhotometry:
 
         return star_groups
 
-    def nstar(self, image, star_groups, *, mask=None):
+    def nstar(self, image, star_groups, *, mask=None, progressbar=None):
         """
         Fit, as appropriate, a compound or single model to the given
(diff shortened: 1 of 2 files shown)


photutils/psf/photometry.py
```diff
@@ -243,14 +243,14 @@ def _make_mask(image, mask):
                               'automatically ignored.', AstropyUserWarning)
         return mask
 
-    def __call__(self, image, *, mask=None, init_guesses=None):
+    def __call__(self, image, *, mask=None, init_guesses=None, **kwargs):
```
larrybradley, line 246: Why not just add the progress bar keyword?

photutils/psf/photometry.py
```diff
@@ -401,7 +404,7 @@ def do_photometry(self, image, *, mask=None, init_guesses=None):
 
         return star_groups
 
-    def nstar(self, image, star_groups, *, mask=None):
+    def nstar(self, image, star_groups, *, mask=None, progressbar=None):
```
larrybradley, line 407: For consistency with with other `photutils` functions, this keyword should be called `progress_bar`.

photutils/psf/photometry.py
```diff
@@ -428,6 +431,11 @@ def nstar(self, image, star_groups, *, mask=None):
             a `True` value indicates the corresponding element of
             ``image`` is masked.
 
+        progressbar : progressbar
+            A progressbar (optional) to show progress over the star groups.
+            Could be an `astropy.utils.console.ProgressBar` or a `tqdm`
+            progressbar.
+
```
larrybradley, line 438: Until just now, I was unaware that astropy even provided a `ProgressBar`.  I think the python community has converged on `tqdm` as the progress bar standard.  I would prefer that this simply be a True/False boolean, and if True would use a `tqdm` progress bar (`tqdm` is an optional dependency).  This makes the API consistent with the `progress_bar` options already available in `photutils` (e.g. see `deblend_sources`, `SourceFinder`, `ESPFBuilder`, `ImageDepth`).  Specifically, I've used the `tqdm.auto` progress bar, which switches from a text-based progress bar for Jupyter notebooks (see, e.g. https://github.com/astropy/photutils/blob/main/photutils/utils/depths.py#L247).

larrybradley (changes requested):
Thanks, @keflavich.  Please see the inline comments.

larrybradley:
This will also need a change log entry.

photutils/psf/photometry.py
keflavich: I didn't know about `tqdm.auto`; I agree that alleviates any need for user-specified progressbars in general.

larrybradley (approved):
Thanks, @keflavich

PR #1443, opened 2022-11-21
Pin setuptools max version

larrybradley:
This PR pins the  `setuptools` max version to workaround https://github.com/numpy/numpy/issues/22623 and https://github.com/pypa/setuptools/issues/3693 until a more permanent solution for `astropy` can be found.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ['setuptools',
+requires = ['setuptools<65.6.0',
             'setuptools_scm',
             'wheel',
             'cython>=0.29.30',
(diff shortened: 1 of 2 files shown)


pllim:
Might want a follow up issue to revert this when upstream figured it out... 😬

saimn:
This should not be needed for photutils. How photutils was affected by the issue ? I only see one build failing before of an `astropy.convolution` import.

larrybradley:
@saimn photutils source dist failed because of astropy: 
 https://github.com/astropy/photutils/actions/runs/3511604458/jobs/5882531510

saimn:
@larrybradley - yes, so pinning in `pyproject.toml` is useless. Adding setuptools to photutils' dependencies can work as a quick temporary fix, but the idea was to wait to bit to see if setuptools and/or numpy are going to fix it. If not it can be fixed in Astropy but with a bit more work (replacing `ctypes` in the convolution module).

Issue #1444, opened 2022-11-21
Revert setuptools max version pin


larrybradley:
The `setuptools` max version pin should be released when a more permanent solution is found to
https://github.com/astropy/astropy/issues/14025

PR #1454, opened 2022-11-28
Remove setuptools pin

larrybradley:
`astropy` has replaced `numpy.ctypeslib` with `cython`, which resolves the convolution input error with setuptools 65.6.0. This PR removes the version pin set in #1443.

Closes: #1444


pllim (approved):
Yay!

PR #1484, opened 2022-12-22
Remove setuptools from install_requires

bnavigator:
Setuptools is not required during runtime.

Remnant of #1443, #1454


larrybradley (approved):
Thanks, @bnavigator.

PR #1543, opened 2023-05-31
Support for measurement uncertainties in PSF fitting

bmorris3:
PSF fitting returns best-fit parameter uncertainties from the astropy modeling framework. The astropy fitters can only take into account measurement uncertainties if the user provides them as `weights`, which currently can't be passed through photutils.

This PR adds `uncertainty` kwargs to PSF photometry methods, and passes those uncertainties as weights to the fitter when available. 

I've added a modified version of the simplest PSF fitting test, which checks that best-fit parameter uncertainties increase as the measurement uncertainties increase.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ General
 New Features
 ^^^^^^^^^^^^
 
+- Propagate measurement uncertainties in PSF fitting. [#1543]
+
 Bug Fixes
 ^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


larrybradley (changes requested):
Overall LGTM -- thanks, @bmorris3!

The python 3.8 CI tests are failing because https://github.com/astropy/astropy/pull/14519 was introduced in astropy 5.3.  Photutils supports astropy >= 5.0 and the 3.8 CI tests use astropy 5.2.2.  You'll need to update the tests comparing the output flux uncertainties only for astropy 5.3+ (e.g., see https://docs.astropy.org/en/stable/api/astropy.utils.introspection.minversion.html).

larrybradley (approved):
Thanks, @bmorris3

PR #1558, opened 2023-06-15
Rewrite PSF photometry classes

larrybradley:
This extremely large PR is a complete rewrite of the PSF photometry classes, adding new classes (the existing classes are unchanged).  I started out refactoring the existing code to make it more performant, but more significant changes were needed.  It was much easier to rewrite (although this was a huge effort) than refactor with many deprecations or backwards-incompatible changes.

The new PSF photometry classes are called `PSFPhotometry` and `IterativePSFPhotometry`.  There are many new features (and fixes) in these classes including:

- At least 20x faster for typical images (500 sources in a 4k x 4k image).  For 5000 sources in a 10k x 10k image, the new `PSFPhotometry` class is 400x faster!
- Error arrays can now be input, which will be used as weights in the fitting. When using astropy 5.3+, these errors will also be propagated to the model fit parameters.
- The source grouper is optional
- The output table is always in source ID order (not group ID order)
- Adds two quality-of-fit metrics to the output table
- Adds more information (columns and metadata) in the output table, including a `flags` column
- Fit warnings are not emitted for each source. A single warning is emitted at the end of fitting.
- Fixes issues with source masking
- The initial parameters tables is more flexible for the x, y, and flux column names
- Supports `NDData` objects
- Supports units
- Allows access to the fitter details (e.g., fit info, parameter covariances)
- Allows access to the finder results
- Adds a local background subtraction option
- no repeated code
- `IterativePSFPhotometry`  uses composition instead of inheritance
- Adds a keyword to control the fitter `maxiters`
- and more...

I plan to deprecate the existing PSF photometry classes, but I'll do that in a separate PR.

A simple notebook (fitting 5000 sources in a 10k x 10k image) is available here: https://gist.github.com/larrybradley/65c0d11c155f3d117eebd30c21d61b31
This takes ~28s on my (old) laptop.  The existing photutils PSF class takes ~3h 10m to perform the same photometry!

Heads up to @bmorris3, @eteq (I don't expect you to review this huge PR! :smile:).


photutils/psf/photometry.py
```diff
@@ -3,1131 +3,1340 @@
...
+                self.grouper = None
 
-        return star_groups
+        if 'local_bkg' not in init_params.colnames:
+            if self.localbkg_estimator is None:
+                local_bkg = np.zeros(len(init_params))
+            else:
+                local_bkg = self.localbkg_estimator(
+                    data, init_params[self._xinit_name],
+                    init_params[self._yinit_name], mask=mask)
```
bmorris3, line 349: Here you supply a `mask`, but it looks like `BackgroundBase.__call__` doesn't expect this: 

https://github.com/astropy/photutils/blob/29867325d7ced75991c13894d2f3a27432445fbe/photutils/background/core.py#L47-L48

bmorris3, line 349:
```suggestion
                local_bkg = self.localbkg_estimator(
                    data, masked=np.any(mask))
```

photutils/psf/groupers.py
```diff
@@ -0,0 +1,106 @@
...
+            The 1D arrays of the x and y centroid coordinates of the
+            sources.
+
+        Returns
+        -------
+        result : 1D int `~numpy.ndarray`
+            A 1D array of the groups, in the same order as the input x
+            and y coordinates.
+        """
+        from sklearn.cluster import DBSCAN
```
bmorris3, line 98: Should this be added to the optional deps?

larrybradley: `scikit-learn` is already an optional dep.

photutils/psf/photometry.py
larrybradley: The `localbkg_estimator` can only be an instance of `LocalBackground`, which takes a`mask`:
 `https://photutils.readthedocs.io/en/latest/api/photutils.background.LocalBackground.html#photutils.background.LocalBackground
 
 `LocalBackground` combines an annulus aperture and a background statistic `BackgroundBase`.

larrybradley: It was in the docstring, but I also just added an explicit check that `localbkg_estimator` is a `LocalBackground` instance or `None`.  Otherwise, it raises a `ValueError`

PR #1563, opened 2023-06-30
Fix for PSF fitting cfit column with out of bounds centroids

bmorris3:
If a PSF fit's central pixel is outside the bounds of the image (`data`), the calculation for the `cfit` column in the photometry result table tries to index `data` out of bounds and fails with an error. This PR checks before it indexes out of bounds, and masks the result in that case.

### DIFF
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -750,7 +750,7 @@ class PSFPhotometry:
                     if cen_in_data:
                         cen_residual = data[ycen, xcen] - model(xcen, ycen)
                     else:
-                        cen_residual = np.ma.masked
+                        cen_residual = np.nan
                 else:
                     # find residual at (xcen, ycen)
                     cen_residual = -residual[cen_idx_]
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -741,8 +741,16 @@ class PSFPhotometry:
                 qfit.append(np.sum(np.abs(residual)) / flux_fit)
 
                 if np.isnan(cen_idx_):
-                    # calculate residual at central pixel
-                    cen_residual = data[ycen, xcen] - model(xcen, ycen)
+                    # calculate residual at central pixel if the central pixel
+                    # is within the bounds of the ``data``, otherwise mask it:
+                    cen_in_data = (
+                        0 <= ycen <= data.shape[0] - 1 and
+                        0 <= xcen <= data.shape[1] - 1
+                    )
+                    if cen_in_data:
+                        cen_residual = data[ycen, xcen] - model(xcen, ycen)
+                    else:
+                        cen_residual = np.ma.masked
                 else:
                     # find residual at (xcen, ycen)
                     cen_residual = -residual[cen_idx_]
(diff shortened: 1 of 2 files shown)


photutils/psf/photometry.py
```diff
@@ -749,8 +749,16 @@ def _calc_fit_metrics(self, data, source_tbl):
...
+                    # calculate residual at central pixel if the central pixel
+                    # is within the bounds of the ``data``, otherwise mask it:
+                    cen_in_data = (
+                        0 <= ycen <= data.shape[0] - 1 and
+                        0 <= xcen <= data.shape[1] - 1
+                    )
+                    if cen_in_data:
+                        cen_residual = data[ycen, xcen] - model(xcen, ycen)
+                    else:
+                        cen_residual = np.ma.masked
```
larrybradley, line 761: My slight preference would be to return `np.nan` here.  And it would also be nice to add a new flag (see `_define_flags`) for this condition:  '32' for center pixel is outside of the image.

larrybradley (commented):
Thanks, @bmorris3!  Could you also add a small regression test for this case?

photutils/psf/tests/test_photometry.py
```diff
@@ -497,3 +498,28 @@ def test_iterative_psf_photometry_inputs():
...
+    fit_shape = (11, 11)
+    psfphot = PSFPhotometry(psf_model, fit_shape,
+                            aperture_radius=10)
+
+    phot = psfphot(data, init_params=Table({'x_init': sources['x_0'], 'y_init': sources['y_0']}))
+
+    # at least one of the best-fit centroids should be
+    # out of the bounds of the dataset, producing a
+    # masked value in the `cfit` column:
+    assert np.any(phot['cfit'].mask)
```
larrybradley, line 525: Should this be `np.any(np.isnan(..))` now?

larrybradley: This should fix the failing test.

photutils/psf/tests/test_photometry.py
```diff
@@ -497,3 +498,28 @@ def test_iterative_psf_photometry_inputs():
...
                                    aperture_radius=4, maxiters=3.14)
+
+
+@pytest.mark.skipif(not HAS_SCIPY, reason='scipy is required')
+def test_out_of_bounds_centroids():
+    sources = Table()
+    sources['id'] = np.arange(4) + 1
+    sources['flux'] = 1
+    sources['x_0'] = [-1.4, 34.5, 14.2, -0.7]
+    sources['y_0'] = [13, -0.2, -1.6, 40]
```
larrybradley, line 509: Can you also add a sources on the top and right edges?  The negative positions would actually not fail for `data[ycen, xcen]` because of negative-index-wraparound (but not what was intended!).

photutils/psf/tests/test_photometry.py
```diff
@@ -497,3 +498,28 @@ def test_iterative_psf_photometry_inputs():
...
+
+    shape = (51, 51)
+    data = make_gaussian_prf_sources_image(shape, sources)
+
+    psf_model = IntegratedGaussianPRF(flux=1, sigma=3.1)
+    fit_shape = (11, 11)
+    psfphot = PSFPhotometry(psf_model, fit_shape,
+                            aperture_radius=10)
+
+    phot = psfphot(data, init_params=Table({'x_init': sources['x_0'], 'y_init': sources['y_0']}))
```
larrybradley, line 520: Also, you can now simply input `init_params=sources` as of https://github.com/astropy/photutils/pull/1566

larrybradley:
This needs a rebase to resolve conflicts.

larrybradley (approved):
Thanks, @bmorris3!

PR #1579, opened 2023-07-11
More progressbar

keflavich:
I wanted to see progress for the pre-fitting aperture photometry too.

I'm not sure I've done things right here - photutils underwent a major refactor since I last dug into the code, which I thought was only a few months ago.

### DIFF
--- a/photutils/aperture/core.py
+++ b/photutils/aperture/core.py
@@ -15,6 +15,7 @@ from astropy.utils import lazyproperty
 
 from photutils.aperture.bounding_box import BoundingBox
 from photutils.utils._wcs_helpers import _pixel_scale_angle_at_skycoord
+from photutils.utils._progress_bars import add_progress_bar
 
 __all__ = ['Aperture', 'SkyAperture', 'PixelAperture']
 


keflavich:
There might be a problem....

```
WARNING: AstropyDeprecationWarning: The get_grouped_psf_model function is deprecated and may be removed in a future version. [photutils.psf.photometry_depr]
WARNING: The fit may be unsuccessful; check fit_info['message'] for more information. [astropy.modeling.fitting]
WARNING: AstropyDeprecationWarning: The subtract_psf function is deprecated and may be removed in a future version. [photutils.psf.photometry_depr]
WARNING: AstropyDeprecationWarning: The _extract_psf_fitting_names function is deprecated and may be removed in a future version. [photutils.psf.utils]
Fit source/group:   2%|███▉
```

larrybradley:
How many apertures are you using to need a progress bar?  Aperture photometry should be very fast.  Are the apertures large?

larrybradley:
> There might be a problem....

I just completely rewrote the PSF photometry classes (and some associated tools).  Among other things, they are significantly faster (by ~1.5-2 orders of magnitude).  See the new PSF photometry docs:  https://photutils.readthedocs.io/en/latest/psf.html

keflavich:
Right, so, I started this on the old codebase before finding the revised versions.  It might be fast enough that this is no longer needed.  I'm running tests now.

I'm looking at roughly 200k stars in one field.

larrybradley:
That's a lot of stars! 👀  The new PSF photometry classes should be dramatically faster.  Let me know if you find issues.

Aperture photometry by itself was already very fast.

keflavich:
yeah, it's the Galactic Center w/JWST.  I'm sure we're hitting the confusion limit in many parts of the image.

I might agree about the aperture photometry in this new version.  The old version, aperture phot was much faster, but still noticeable.

keflavich:
This isn't really the right place for a bug report, but since I'm trying things on the fly - and am far from being able to make an MWE - I'll note here: my first attempt with the new code ran out of memory about 7% through the fitting process.  I'm going to retry with grouping disabled.

keflavich:
I think we'll abandon this PR, but I have an open issue with progressbars that may be addressable: I run `photutils` in batch jobs, in which I want to have some way to track progress.  However, progressbars are not a great way to do this: they are meant for interactive use, since they use the same screen real estate, while when logging to file they spam them full of redundant messages.

It would be helpful to have a more configurable approach to progressbars.  As is, I could monkeypatch the `photutils.utils._progress_bars.add_progress_bar` to add a different type of progress bar, or monkeypatch tqdm to achieve the same, but that's not a very nice approach.

Given the current progress bar infrastructure, I'm not sure what is easiest.  My previous approach was to allow a callable to be passed to the `progress_bar` arguments, but now those only accept booleans.  Would a refactor to allow a tqdm-like object be a good approach?

larrybradley:
I'd really like to keep the current simple boolean implementation for progress bars -- most users have never heard of `tqdm` and will not want to have to deal with inputting a `tdqm` object.  The current implementation of having both text-based and notebook (ipywidgets-based) tqdm progress bars is also a bit tricky.

However, I'm open to adding something else for your use case.  I don't know what you mean though.  What other kind of progress bar do you have in mind? Could you provide an minimal example so I can get an idea?

larrybradley:
It could be `progress_bar = True | False | callable`, but what would `callable` do/look like?

keflavich:
The use case is logging to file and having some sense of how far along the code was when it crashed - this is crucial for debugging on large photometry jobs, like those I've been pushing recently.

A very simple example might be to pass in a `tqdm` object with `miniters=100`, e.g.:
```python
from tqdm import tqdm
class tqdm_100(tqdm):
     def __init__(self, *args, **kwargs):
        return super().__init__(*args, miniters=100, **kwargs)
```
(note that I haven't tested this)

basically with the idea that there would be less frequent dumps to file.

Another approach might be to make the progressbar be a simpler iterable wrapper:
```python
def progressbar(iterable):
     for item in iterable:
        print(".", end=None, flush=True)
        yield item
     print("\n", flush=True)
```

These are unsophisticated approaches that can significantly reduce the size of log files.

larrybradley:
How exactly are you logging `tqdm` output?  Did you write a custom `logging` handler for `tqdm` (because it shouldn't go to a logger)?  If instead you are using shell redirection, `tqdm` writes to `stderr`, so it should only appear in a terminal (not log file) unless you explicitly request it.  Can you provide a MWE of exactly how you are logging `tqdm` output?

keflavich:
I'm running slurm batch jobs on a supercomputer.  Both stdout and stderr are redirected to log files, which is necessary since there is no live terminal frontend for supercomputer jobs.

larrybradley:
Do stdout and stderr (hopefully) go to separate log files (e.g., `1 > log.txt 2 > err.txt` ?  Is the issue that the stderr log file is too large?  It should be a relatively small text file with (hopefully) only `tqdm` output that you can delete after the job.

keflavich:
No, though I could do that.  It actually comes out OK right now if I `cat` the file, since it prints the whole thing to terminal (it replays the whole progressbar at once), but it would be nice to have some control over how many times the print happens.  Even if one gets redirected to stderr, we're still flooding the log file with unhelpful extra text.

That said... let's punt this for now, it's not _that_ urgent.

larrybradley:
Instead of `cat`, you can use `tail`.  In a terminal you could even use `tail -f errlog.txt` to watch the progress bar in real time.

PR #1621, opened 2023-09-12
Fixes for NumPy 2.0 changes

### LINKED ISSUE #1620: No module named 'numpy.lib.index_tricks' (numpy 2.0)
Saw it this morning over at Jdaviz using numpy nightly wheel.

```
.../photutils/__init__.py:16: in <module>
    from . import background
.../photutils/background/__init__.py:7: in <module>
    from .background_2d import *  # noqa: F401, F403
.../photutils/background/background_2d.py:15: in <module>
    from numpy.lib.index_tricks import index_exp
E   ModuleNotFoundError: No module named 'numpy.lib.index_tricks'
```

https://github.com/astropy/photutils/blob/5d8795814b07654541eaa7ddde9031c481eb246f/photutils/background/background_2d.py#L15C13-L15C13

Hard to tell from https://github.com/numpy/numpy/pull/24581 but looks like all you need to do is use `numpy.index_exp` directly? Maybe it is even backward compatible.

larrybradley:
Closes #1620

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -79,6 +79,8 @@ filterwarnings =
     ignore:numpy\.ndarray size changed:RuntimeWarning
     # numpy 1.24 deprecation from skimage (can remove for skimage >= 0.20)
     ignore:`np.bool8` is a deprecated alias for `np.bool_`:DeprecationWarning
+    # TEMP: remove when matplotlib nightly fixes the row_stack deprecations
+    ignore:`row_stack` alias is deprecated:DeprecationWarning
 
 [coverage:run]
 omit =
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -79,6 +79,8 @@ filterwarnings =
     ignore:numpy\.ndarray size changed:RuntimeWarning
     # numpy 1.24 deprecation from skimage (can remove for skimage >= 0.20)
     ignore:`np.bool8` is a deprecated alias for `np.bool_`:DeprecationWarning
+    # TEMP: remove when matplotlib nightly fixes the row_stack deprecations
+    ignore:`row_stack` alias is deprecated:DeprecationWarning
```
pllim, line 83: Hmm why didn't astropy stumble on this? 🤔

larrybradley: It's actually fixed in mpl main, but the `devdeps` test picks up the rc version, which does *not* have the fix.

pllim: Ah, right!

But then again, numpy is also in RC, so it would cancel out for a while. 🤪

PR #1636, opened 2023-10-03
Fix source deblending if contrast = 1

larrybradley:
This PR fixes an issue where ``deblend_sources`` and ``SourceFinder`` would raise an error if the ``contrast`` keyword was set to 1 (meaning no deblending).


larrybradley:
Failing devdeps tests are due to NumPy 2.0dev deprecations in upstream packages.

PR #1667, opened 2023-11-21
TST: Fix devdeps

pllim:
devdeps job is failing

xref https://github.com/astropy/astropy/issues/15638

Since this repo has C-extension, might also need something like https://github.com/spacetelescope/synphot_refactor/pull/368

### Blocked by

* https://github.com/scikit-image/scikit-image/issues/7249

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ requires = [
     'setuptools>=61.2',
     'setuptools_scm>=6.2',
     'cython>=3.0.0,<3.1.0',
-    'numpy',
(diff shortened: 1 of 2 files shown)


pllim:
Needs to wait for scikit-image to catch up. Same problem as https://github.com/spacetelescope/acstools/pull/192 and https://github.com/spacetelescope/jdaviz/pull/2572

```
ImportError: numpy.core.multiarray failed to import
It seems that scikit-image has not been built correctly.
```

Issue #1780, opened 2024-06-18
DAOPhot is finding sources that are out-of-bounds


keflavich:
A recent change https://github.com/astropy/photutils/commit/9994d0bfca061ec06ebeec03519d2f3ad48e69ef resulted in this error:

```python
  File /blue/adamginsburg/adamginsburg/jwst/brick/analysis/crowdsource_catalogs_long.py:832 in do_photometry_step
    result = phot(np.nan_to_num(data))
  File /blue/adamginsburg/adamginsburg/repos/photutils/photutils/psf/photometry.py:1238 in __call__
    fit_inputs = self._prepare_fit_inputs(data, mask=mask, error=error,
  File /blue/adamginsburg/adamginsburg/repos/photutils/photutils/psf/photometry.py:1099 in _prepare_fit_inputs
    self._check_init_positions(init_params, data.shape)
  File /blue/adamginsburg/adamginsburg/repos/photutils/photutils/psf/photometry.py:657 in _check_init_positions
    raise ValueError('Some of the sources have no overlap with the '
ValueError: Some of the sources have no overlap with the data. Check the initial source positions or increase the fit_shape.
```

The call is fairly simple:
```python
        phot = PSFPhotometry(finder=daofind_tuned,
                             localbkg_estimator=LocalBackground(5, 15),
                             psf_model=dao_psf_model,
                             fitter=LevMarLSQFitter(),
                             fit_shape=(5, 5),
                             aperture_radius=2*fwhm_pix,
                             progress_bar=True,
                            )
       phot(data)
```

I can try increasing the `fit_shape` here, but I don't understand the logic. 

Also, here's the catalog:
```python
ipdb> init_params['x_init'].min()
5.229948988964448
ipdb> init_params['x_init'].max()
5686.278841347583
ipdb> init_params['y_init'].max()
2329.4421707209644
ipdb> init_params['y_init'].min()
4.140421280588503
ipdb> shape
(2334, 5692)
```

I'm not sure why this even failed, since it should've been in bounds already....

...it's possible that this failure is not caused by the recent change, since the logic seems to have gone unchanged, but it is definitely a regression since a photutils version from last year.

keflavich:
The problem is this line: x, y are swapped
https://github.com/astropy/photutils/blame/4b179c59b083692b07bc483b6af22c1adba7ac93/photutils/psf/photometry.py#L644

larrybradley:
Fixed in #1781

PR #1781, opened 2024-06-19
swap xy in validity check

keflavich:
Fix for #1780


larrybradley (approved):
Yikes!  Thanks, @keflavich!

PR #1783, opened 2024-06-20
Make flux_range a keyword in make_model_params

larrybradley:
This PR also adds examples in the `datasets` subpackage.

### DIFF
--- a/photutils/datasets/sources.py
+++ b/photutils/datasets/sources.py
@@ -17,8 +17,8 @@ __all__ = ['make_model_params', 'make_random_models_table',
            'make_random_gaussians_table']
 
 
-def make_model_params(shape, n_sources, flux_range, *, min_separation=1,
-                      border_size=(0, 0), seed=0):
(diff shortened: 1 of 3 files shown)


photutils/datasets/sources.py
```diff
@@ -37,7 +39,7 @@ def make_model_params(shape, n_sources, flux_range, *, min_separation=1,
         given ``shape`` and therefore the number of sources generated
         may be less than ``n_sources``.
 
-    flux_range : 2-tuple
+    flux_range : 2-tuple, optional
```
VascoSch92, line 42: actually it is not optional. It has a default value :-) optional means that the default value is None

larrybradley: @VascoSch92 "Optional" means that the user does not need to specify a value (e.g., all keywords are "optional" in numpydoc style).  Please see the `numpydoc` style doc:  https://numpydoc.readthedocs.io/en/latest/format.html

Issue #1799, opened 2024-06-29
Reject negative fluxes?


keflavich:
Fitted PSF photometry fluxes can go negative.  

Can we include a mechanism to reject them using public methods?

Right now, you can reject them, but it requires modifying a private method.  You need this if you're interested in using the residual image or model image.  Example:

```python
phot = PSFPhotometry()
result = phot(data)
bad = result['flux_fit'] <= 0
result = result[~bad]
phot._fit_models = [mod for mod, ok in zip(phot._fit_models, ~bad) if ok]
```

keflavich:
I'm also not sure that works, since [`make_model_image`](https://github.com/astropy/photutils/blob/a42123c547c0b75ac91e0065a4f5107d0d90e811/photutils/psf/photometry.py#L72) uses `_fit_model_params`, but that doesn't exist for the `PSFPhotometry` instances I'm playing with.

keflavich:
with latest photutils, should be:
```python
phot = PSFPhotometry()
result = phot(data)
bad = result['flux_fit'] <= 0
result = result[~bad]
phot._fit_model_params = phot._fit_model_params[~bad]
```

keflavich:
Quoting @larrybradley from https://github.com/astropy/photutils/pull/1800#issuecomment-2200314737:

> To exclude negative peaks you should bound the flux parameter on the input psf_model, e.g, psf_model.flux.min = 0.

PR #1800, opened 2024-07-01
Add option to exclude negative peaks from residual/model image

keflavich:
Solves #1799 by making negative peaks excludable.

It's difficult to impossible for a user to do this now, since the construction of the list of fits for IterativePSFPhotometry is done within `make_model_image`.

### DIFF
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -164,8 +164,8 @@ class ModelImageMixin:
                 exclude_negative_peaks=exclude_negative_peaks)
         else:
             residual = self.make_model_image(data.shape, psf_shape,
-                                             include_localbkg=include_localbkg,
-                                             exclude_negative_peaks=exclude_negative_peaks)
+                include_localbkg=include_localbkg,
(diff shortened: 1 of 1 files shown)


larrybradley:
@keflavich To exclude negative peaks you should bound the `flux` parameter on the input `psf_model`, e.g, `psf_model.flux.min = 0`.

I've considered automatically excluding negative flux results, but I've only seen that occur with `IterativePSFPhotometry` where something has gone wrong (i.e., the source doesn't really exist).  I wasn't sure whether removing them is the correct choice, since it gives the user an indication that something went wrong -- also because the user can bound the flux parameter themselves if they only want positive flux.

In https://github.com/astropy/photutils/pull/1778, I do automatically exclude fit (x, y) positions that out-of-bound during the `IterativePSFPhtometry` iterations because it will crash without it.  The (x, y) model parameters could also be bound in the input `psf_model` to the image shape.

keflavich:
Thanks.  That makes sense as a way to prevent the fit from ever going negative.

I am seeing this show up even with "basic" PSFPhotometry.  I think it has to do with a structured background.

PR #1804, opened 2024-07-09
Add description of bound parameter to PSF docs

### LINKED ISSUE #1799: Reject negative fluxes?

larrybradley:
Closes #1799.

CC: @keflavich


larrybradley:
Well, they can be bounded (e.g., to an image shape), it's just that the bounds are global for all sources.   I'll add that for now. 
 I'm working on the x/y bounds PR now.

PR #1805, opened 2024-07-09
Add xy_bounds keyword for PSF photometry

### LINKED ISSUE #1801: Limit offsets when fitting
As noted in #1799, it is possible to limit the flux to be positive by setting the model flux min to be zero:
```python
    dao_psf_model.flux.min = 0
```

which is useful for rejecting negative sources during fitting.

I'd like a mechanism to do the same for `x_0` and `y_0`, but restricting those parameters doesn't work since they need to be set to the local x,y value.  Instead, the limits need to be set independently for each source.  I think that could be accommodated in `make_psf_model`.

One use case is very crowded fields:  I've seen DAOFind-found peaks close enough that they jump within the 5x5 fit zone onto another star's peak.  

A workaround I'll try (but expect to have limited success with) is grouping.

larrybradley:
This PR adds a new ``xy_bounds`` keyword to ``PSFPhotometry`` and ``IterativePSFPhotometry`` to allow one to bound the x and y model parameters during the fitting.  If any fitted x or y value is at the bounded value, then a new flag value of 32 is added to the bitwise `flags` column.

Closes #1801 

CC: @keflavich

### DIFF
--- a/photutils/psf/photometry.py
+++ b/photutils/psf/photometry.py
@@ -226,6 +226,14 @@ class PSFPhotometry(ModelImageMixin):
         is not converging for sources (e.g., the output ``flags`` value
         contains 8).
 
+    xy_bounds : `None`, float, or 2-tuple of float, optional
+        The maximum distance in pixels that a fitted source can be from
+        the initial (x, y) position. If a single float, then the same
+        maximum distance is used for both x and y. If a 2-tuple of
+        floats, then the distances are in ``(x, y)`` order. If `None`,
+        then no bounds are applied. Either value can also be `None` to
+        indicate no bound in that direction.
+
     localbkg_estimator : `~photutils.background.LocalBackground` or `None`, optional
         The object used to estimate the local background around each
         source. If `None`, then no local background is subtracted. The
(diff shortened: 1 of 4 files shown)


photutils/psf/photometry.py
```diff
@@ -289,20 +297,21 @@ class for more information.
...
                  fitter=LevMarLSQFitter(), fitter_maxiters=100,
-                 localbkg_estimator=None, aperture_radius=None,
+                 xy_bounds=None, localbkg_estimator=None, aperture_radius=None,
                  progress_bar=False):
 
         self._param_maps = self._define_param_maps(psf_model)
         self.psf_model = _validate_psf_model(psf_model)
 
-        self.fit_shape = as_pair('fit_shape', fit_shape, lower_bound=(0, 1),
+        self.fit_shape = as_pair('fit_shape', fit_shape, lower_bound=(1, 0),
```
keflavich, line 306: Is this change intentional?  I'm not sure what it does, but it seems smuggled in

keflavich (approved):

photutils/psf/photometry.py
larrybradley: It's a tiny "fix" that I noticed during testing.  All this change does is change the error message if one were to input `fit_shape = 0`.  It could be a separate PR, but it's tiny.....

PR #1834, opened 2024-08-09
Add publishing of developer wheels to anaconda.org

astrofrog:
This will upload the wheels here: https://anaconda.org/astropy as for reproject and astropy-healpix

I've set up the token in the repository settings under ANACONDA_TOKEN. If there are permission issues when trying to upload a wheel, I can try and update the token scope.

PR #1835, opened 2024-08-12
DOC: Mention nightly wheel in dev install

pllim:
Follow-up of #1834

p.s. The `[all]` actually does not work right now because of `bottleneck` but fixing that is out of scope here.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -141,7 +141,7 @@ the "-e" option::
 Alternatively, `pip`_ can be used to retrieve and install the
 latest development pre-built wheel::
 
-    pip install --upgrade --index-url https://pypi.anaconda.org/astropy/simple photutils[all] --pre
+    pip install --upgrade --extra-index-url https://pypi.anaconda.org/astropy/simple photutils[all] --pre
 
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -138,11 +138,10 @@ the "-e" option::
...
     pip install -e ".[all]"
 
-Alternatively, `pip`_ can be used to retrieve, build, and install the
-latest development version from `GitHub`_::
-
-    pip install "photutils[all] @ git+https://github.com/astropy/photutils.git"
+Alternatively, `pip`_ can be used to retrieve and install the
+latest development pre-built wheel::
 
+    pip install --upgrade --index-url https://pypi.anaconda.org/astropy/simple photutils[all] --pre
```
larrybradley, line 144:
```suggestion
    pip install --upgrade --extra-index-url https://pypi.anaconda.org/astropy/simple photutils[all] --pre
```

larrybradley (changes requested):

larrybradley (approved):
Thanks!

Issue #1808, opened 2024-07-10
Excessive memory usage when using source grouping


keflavich:
I have an MWE that fails reliably now.  It's not all _that_ minimal, but minimal enough I hope.

```python
import numpy as np
from photutils.psf import PSFPhotometry, IterativePSFPhotometry, SourceGrouper
from photutils.detection import DAOStarFinder
from photutils.background import LocalBackground

import scipy.ndimage
from astropy.io import fits
from astropy.modeling.fitting import LevMarLSQFitter

from webbpsf.utils import to_griddedpsfmodel

basepath = '/blue/adamginsburg/adamginsburg/jwst/brick/'

im1 = fits.open(f'{basepath}/analysis/MWE_example.fits')
obsdate = im1[0].header['DATE-OBS']
module = 'merged'
data = im1['SCI'].data[:400, :400]
err = im1['ERR'].data[:400, :400]

weight = err**-1
maxweight = np.percentile(weight[np.isfinite(weight)], 95)
minweight = np.percentile(weight[np.isfinite(weight)], 5)
badweight = np.percentile(weight[np.isfinite(weight)], 1)
weight[err < 1e-5] = 0
weight[np.isnan(weight)] = 0
bad = np.isnan(weight) | (data == 0) | np.isnan(data) | (weight == 0) | (err == 0) | (data < 1e-5)

mask = bad

mask = scipy.ndimage.binary_dilation(scipy.ndimage.binary_erosion(mask, iterations=1), iterations=1)
mask = scipy.ndimage.binary_erosion(scipy.ndimage.binary_dilation(mask, iterations=1), iterations=1)

filtered_errest = np.nanmedian(err)

fwhm_pix = 2.302

daofind_tuned = DAOStarFinder(threshold=5 * filtered_errest,
                              fwhm=fwhm_pix, roundhi=1.0, roundlo=-1.0,
                              sharplo=0.30, sharphi=1.40,
                              exclude_border=True
                              )
grouper = SourceGrouper(2 * fwhm_pix)


oversample = 1
proposal_id = 1182
field = '004'
filtername = 'f444w'

blur_ = '_blur'
dao_psf_model = to_griddedpsfmodel(f'{basepath}/psfs/{filtername.upper()}_{proposal_id}_{field}_merged_PSFgrid_oversample{oversample}{blur_}.fits')
dao_psf_model.flux.min = 0

print("Initializing IterativePSFPhotometry")
[...]

PR #1856, opened 2024-08-29
Add notes about excessive memory usage for large star groups

larrybradley:
See #1808.

### DIFF
--- a/docs/psf.rst
+++ b/docs/psf.rst
@@ -750,12 +750,13 @@ stars in each group were simultaneously fit.
       9        6          2
      10        7          1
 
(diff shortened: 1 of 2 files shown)


keflavich:
@larrybradley could we make this threshold:
https://github.com/larrybradley/photutils/blob/5a4330dacae09c09a4982f0aa9bd420749d36783/photutils/psf/photometry.py#L1237
a configuration item? I've been running into OOM issues again that I think are caused by this.  I have ~10,000 sources, but I'm using very large (>50 pixel) footprints, so I think even a small amount of grouping can blow up the memory use;  I'd like to be able to get warnings for N>5, for example.

PR #1875, opened 2024-09-12
Fix weekly CI

larrybradley:


larrybradley:
The ``aarch64`` job times out after 6 hours, but the previously-failing test now passes.

xref: https://github.com/astropy/astropy/issues/16565

Issue #1895, opened 2024-09-25
DOC: Replace LevMarLSQFitter with something else?


pllim:
As per this upstream PR:

* https://github.com/astropy/astropy/pull/16983

PR #1899, opened 2024-09-27
Remove LevMarLSQFitter

larrybradley:
This PR changes the default Astropy fitter for ``PSFPhotometry``, ``IterativePSFPhotometry``, and ``EPSFFitter``  from ``LevMarLSQFitter`` to ``TRFLSQFitter``.

``LevMarLSQFitter`` uses the Levenberg-Marquardt algorithm via the SciPy legacy function ``scipy.optimize.leastsq``, which is no longer recommended. This fitter supports parameter bounds using an unsophisticated min/max condition where parameters that are out of bounds are simply reset to the min or max of the bounds during each step. This can cause parameters to stick to one of the bounds during the fitting process if the parameter gets close to the bound. If needed, this fitter can still be used by explicitly setting the fitter in the ``PSFPhotometry``, ``IterativePSFPhotometry``, and ``EPSFFitter`` classes.

The fitter used in ``RadialProfile`` to fit the profile with a Gaussian was also changed from ``LevMarLSQFitter`` to ``TRFLSQFitter``.

The fitter used in ``centroid_1dg`` and ``centroid_2dg`` was changed from ``LevMarLSQFitter`` to ``LMLSQFitter``.

See https://github.com/astropy/astropy/pull/16983 for more information.

Closes: #1895

### DIFF
--- a/photutils/centroids/gaussian.py
+++ b/photutils/centroids/gaussian.py
@@ -6,7 +6,7 @@ The module contains tools for centroiding sources using Gaussians.
 import warnings
 
 import numpy as np
-from astropy.modeling.fitting import LevMarLSQFitter
+from astropy.modeling.fitting import LMLSQFitter
 from astropy.modeling.models import Gaussian1D, Gaussian2D
 from astropy.utils.exceptions import AstropyUserWarning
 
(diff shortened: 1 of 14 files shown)


photutils/centroids/gaussian.py
```diff
@@ -105,11 +105,13 @@ def centroid_1dg(data, error=None, mask=None):
 
     xy_data = [np.ma.sum(data, axis=i).data for i in (0, 1)]
 
+    # would need to use a different fitter if parameter bounds are used
+    fitter = LMLSQFitter()
```
astrofrog, line 109: @larrybradley - gaussians have bounds enabled by default (on stddev) so if you are fitting a Gaussian I think it would be best to use TRFLSQFitter or another one, not ``LMLSQFitter``

larrybradley: @astrofrog Are you sure about that?  Where are the bounds defined?  I don't see any bounds defined in `functional_models`?  Do you mean constrained to be postive?  I still get negative stddev's sometimes.....

larrybradley: I've also noticed that `TRFLSQFitter` is about 4-5% slower than `LevMarLSQFitter` even without any bounds.

astrofrog: @larrybradley yes the fact the stddev is required to be positive - it's formally treated as a bound.

astrofrog: Do you see negative stddev values with LMLSQ or TRFLSQ or both?

PR #1916, opened 2024-10-08
Temporarily disable docformatter pre-commit hook

larrybradley:
`pre-commit` 4.0.0 removed `language: python_venv`, which breaks the `docformatter` hook.  See https://github.com/PyCQA/docformatter/pull/287


larrybradley:
RTD is stuck, so merging.

PR #1924, opened 2024-10-11
Improve serial deblending performance

larrybradley:
This PR improves the performance and reduces the memory usage of serial (nproc=1) source deblending.


larrybradley:
The failing test is unrelated and due to astropy-dev changes in modeling.

PR #1926, opened 2024-10-14
Improve accuracy of progress bar for multiprocessing source deblending

larrybradley:
This PR also uses `concurrent.futures.ProcessPoolExecutor` and gathers worker tasks as they are completed instead of sequentially.


larrybradley:
The devdeps test failures are unrelated and will be addressed in a separate PR.

Issue #1959, opened 2024-11-23
When WCS provided, should also return center in sky coordinates


pllim:
~Currently, even if WCS is provided for sky aperture, center is only returned in pixel coordinates.~ Currently if WCS is provided but aperture is in pixel coordinates, sky center is not returned.

Desired outcome: A new column named `sky_center` with center as `SkyCoord` whenever `wcs` is provided.

1. If aperture is given in pixel coordinate, if `wcs` is provided, a column named `sky_center` would exist. If not provided, it is not created.
2. If aperture is given in sky coordinate, `sky_center` would exist because `wcs` is required. (This is already the case.)

Motivation: When going through photometry examples in astropy workshop materials, the example is manually inserting this column. I feel like this column can be a natural outcome of aperture photometry from `photutils` itself.

pllim:
I dunno what happened earlier but when I reran an example with sky aperture, I now notice that photutils does return `sky_center`. I updated my original post above.

PR #1965, opened 2024-11-27
Always include sky_center column if wcs is input to aperture_photometry

larrybradley:
With this PR, the ``aperture_photometry`` output table will now include a ``sky_center`` column if ``wcs`` is input, even if the input aperture is not a sky aperture.

Closes: #1959


pllim:
Thank you!

PR #1972, opened 2024-12-19
DOC: Fix reference for segmentation module

bnavigator:
The docs point to the wrong module.

See https://photutils.readthedocs.io/en/2.0.2/reference/segmentation_api.html


larrybradley (changes requested):
Thanks!  

The RTD build is failing because the over- and underlines do not match the title length.  Can you fix those?
https://readthedocs.org/projects/photutils/builds/26642152/

larrybradley (approved):
Thanks!

PR #1976, opened 2024-12-25
Remove duplicate reference for segmentation module

### LINKED ISSUE #1975: Duplicate references for the segmentation module are present
Image

Two references for the segmentation module appear in the user guide.

devesh-aggarwal:
resolves #1975


larrybradley:
This was intentional.  Image segmentation is a category by itself, but it also includes general source detection and extraction, which differs from the point-like source `detection` submodule.  As part of #1982, I updated the section titles to clarify this distinction.

PR #1978, opened 2024-12-30
Include interval ends for sharpless, roundness, and peakmax

### LINKED ISSUE #1977: Sources that match sharplo(hi) and roundlo(hi) limits should be included
Currently star finding code filters out sources that have sharpness or roundness match one of the limits for these parameter, see https://github.com/astropy/photutils/blob/c15b004be4ecd1256f8a8db8c9c7a52e795fe3a8/photutils/detection/irafstarfinder.py#L580-L583

I believe the strict inequalities in the code above should be replaced with `>=` and `<=`.

Here is an example of a source that should be detected:

```python
In [1]: import numpy as np

In [2]: from photutils import detection

In [3]: d = np.zeros((200, 200))

In [4]: x = 121
   ...: y = 98
   ...: d[y-1:y+2, x-1:x+2] = [
   ...:     [0.1, 0.6, 0.1],
   ...:     [0.6, 0.8, 0.6],
   ...:     [0.1, 0.6, 0.1],
   ...: ]

In [5]: f = detection.IRAFStarFinder(0, 2.5, roundlo=0)

In [6]: f.find_stars(d)
WARNING: NoDetectionsWarning: Sources were found, but none pass the sharpness, roundness, or peakmax criteria [photutils.detection.irafstarfinder]

In [7]: f = detection.IRAFStarFinder(0, 2.5, roundlo=-1e-12)

In [8]: f.find_stars(d)
Out[8]:
<QTable length=1>
  id  xcentroid ycentroid        fwhm            sharpness      roundness    pa    npix   peak         flux                mag
int64  float64   float64       float64            float64        float64  float64 int64 float64      float64             float64
----- --------- --------- ------------------ ------------------ --------- ------- ----- ------- ------------------ -------------------
    1     121.0      98.0 1.5698800300206328 0.6279520120082531       0.0     0.0     9     0.8 3.5999999999999996 -1.3907562519182182
```
That is, perfectly round stars are excluded by default.

mcara:
Closes #1977

This PR modifies selection of sources by shapness, roundness, and peakmax to include the interval ends, i.e., that sources that have roundness 0.0 are selected instead of being rejected when `sharplo=0.0`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -51,6 +51,11 @@ API Changes
     different along the x and y edges if the kernel shape is rectangular.
     [#1957]
 
+  - Detected sources that match interval ends for sharpness, roundness, and
+    maximum peak values (``sharplo``, ``sharphi``, ``roundlo``, ``roundhi``,
+    and ``peakmax``) are now included in the returned table of detected
+    sources by ``DAOStarFinder`` and ``IRAFStarFinder``. [#1978]
+
 - ``photutils.morphology``
 
   - The ``gini`` function now returns zero instead of NaN if the
--- a/photutils/detection/daofinder.py
+++ b/photutils/detection/daofinder.py
@@ -740,13 +740,13 @@ class _DAOStarFinderCatalog:
 
         # filter based on sharpness, roundness, and peakmax
         mask = ((newcat.sharpness > newcat.sharplo)
-                & (newcat.sharpness < newcat.sharphi)
-                & (newcat.roundness1 > newcat.roundlo)
-                & (newcat.roundness1 < newcat.roundhi)
-                & (newcat.roundness2 > newcat.roundlo)
-                & (newcat.roundness2 < newcat.roundhi))
+                & (newcat.sharpness <= newcat.sharphi)
+                & (newcat.roundness1 >= newcat.roundlo)
+                & (newcat.roundness1 <= newcat.roundhi)
+                & (newcat.roundness2 >= newcat.roundlo)
+                & (newcat.roundness2 <= newcat.roundhi))
         if newcat.peakmax is not None:
-            mask &= (newcat.peak < newcat.peakmax)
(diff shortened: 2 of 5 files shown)


larrybradley:
@mcara Do you recall what the old IRAF routines use?  This is a breaking change, but I suppose *could* be classified as a bug.

mcara:
Sorry, I replied in the issue: https://github.com/astropy/photutils/issues/1977#issuecomment-2568689761

mcara:
> This is a breaking change...

This could break some idealized tests, just like the one I am using in https://github.com/astropy/photutils/issues/1977 but in practice, what is the chance to have a perfectly round star (or a star with ellipticity precisely 2) in real observations? So, I believe this PR will have a minor effect, except for helping avoid this kind of tricks: https://github.com/spacetelescope/jwst/blob/6d0c2d50c72d32f9a16bf30264d3528bff53e36e/jwst/tweakreg/tests/test_tweakreg.py#L99 which seem unnatural given that the lower limit for roundness is 0 (it cannot be negative in the original IRAF code) yet I _have to_ set this limit to a negative value in order to include 0.

CHANGES.rst
```diff
@@ -43,6 +43,11 @@ API Changes
     different along the x and y edges if the kernel shape is rectangular.
     [#1957]
 
+  - Detected sources that match interval ends for sharpnes, roundness, and
```
larrybradley, line 46:
```suggestion
  - Detected sources that match interval ends for sharpness, roundness, and
```

larrybradley (changes requested):
LGTM.  The old IRAF tools also allowed values at the limits.

Could you also please add a test for `DAOStarFinder`?

larrybradley:
This also needs to be rebased due to a conflict in the changelog.

larrybradley (approved):
Thanks, @mcara.  The test failure is unrelated to this PR.

PR #1993, opened 2025-01-06
Remove pixel units from (x/y)center columns in aperture_photometry

### LINKED ISSUE #1970: ApertureStats xcentroid and ycentroid should have unit of pix
@ejeschke pointed out that ApertureStats `xcentroid` and `ycentroid` properties are unitless, which is inconsistent with `x_center` and `y_center` from `aperture_photometry` that have unit of `u.pix`. Was this intentional or an oversight? Would be nice if `xcentroid` and `ycentroid` have unit of `u.pix`. Thanks!

larrybradley:
This PR removes the pixel units from the xcenter and ycenter columns in the table output from `aperture_photometry`.  The units were removed to be consistent with all other tools in `photutils`, none of which have `u.pixel` units for x and y columns.  This inconsistency was reported in #1970.

Closes #1970


pllim:
Thanks! Will have to see how this affects Jdaviz downstream. [🐱](https://jira.stsci.edu/browse/JDAT-5058)

PR #1997, opened 2025-01-22
Use the free ARM64 test runner

larrybradley:
CC: @pllim 

The ARM64 test passed and it's very fast, so I moved it from the weekly CI to the regular CI tests.


pllim (approved):
🚀

PR #2008, opened 2025-02-17
Always return ellipse/rectangle pixel aperture theta as an angular Quantity

larrybradley:
The ``theta`` attribute of ``EllipticalAperture``, ``EllipticalAnnulus``, ``RectangularAperture``, and ``RectangularAnnulus`` is now always returned as an angular ``Quantity``.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,13 @@ Bug Fixes
 API Changes
 ^^^^^^^^^^^
 
+- ``photutils.aperture``
+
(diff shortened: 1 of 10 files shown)


pllim:
This breaks Jdaviz test assumption. FYI.

larrybradley:
Is this only for a test?  Where else are you using the `theta` attribute??

pllim:
I thought it was only the test, but turns out I have to adjust our internal translators too. Please see https://github.com/spacetelescope/jdaviz/pull/3444

PR #2009, opened 2025-02-17
Add aperture_to_region function

larrybradley:
This PR adds an ``aperture_to_region`` function to convert an Aperture object to an astropy ``Region`` or ``Regions`` object.

``Region`` and ``Regions`` objects can be serialize to region files in various formats.


pllim:
[🐱](https://jira.stsci.edu/browse/JDAT-5148)

PR #2017, opened 2025-03-02
Fix error raised due to naming scheme for NIRCam SW STDPSFs

avapolzin:
The filter name and detector ID are swapped in Jay Anderson's short wavelength NIRCam STDPSFs. This is a quick bug fix to swap the order of the arguments after checking that the STDPSFs are being pulled from the STScI site and are for the NIRCam short wavelength channel.

As a test, the following code, which previously would have raised an error (`ValueError: Unknown detector F115W.`) should now work:

```
from photutils.psf import GriddedPSFModel
url = 'https://www.stsci.edu/~jayander/JWST1PASS/LIB/PSFs/STDPSFs/NIRCam/SWC/F115W/STDPSF_F115W_NRCA2.fits'
det = 2
model = GriddedPSFModel.read(filename = url, detector_id = det, format= 'stdpsf')
```

I will note that this fix is limited to users who provide the filename via URL, not those who provide a path to a locally downloaded STDPSF file. This was an explicit condition on swapping the detector and filter values to prevent causing issues for users who are providing files with the traditional naming scheme.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,11 @@ Bug Fixes
   - Fixed an issue with the initial Gaussian theta units in
     ``centroid_2dg``. [#2013]
 
+- ``photutils.psf``
+
+  - Adapt ``psf.GriddedPSFModel.read()`` to the changed STDPSF file name
+    convention for NIRCam's short wavelength channel. [#2017]
+
 API Changes
 ^^^^^^^^^^^
 
--- a/photutils/psf/model_io.py
+++ b/photutils/psf/model_io.py
@@ -373,6 +373,10 @@ def _get_metadata(filename, detector_id):
         return None  # filename from astropy download_file
 
     detector, filter_name = parts[1:3]
+    if (inst_det[1] == 'NRCSW') and ('~jayander' in filename.split('/')): 
+        #Fix bug with the Jay Anderson NIRCam short wavelength STDPSFs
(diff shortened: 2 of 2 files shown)


larrybradley:
Hi, @avapolzin .  Many thanks for reporting this and sorry about the delay!

Regarding the trailing whitespace, it is located on line 379 in `model_io.py`.  The line appears to be blank, but it actually contains whitespace.  In the future, you can install and run `pre-commit` locally (`pip install pre-commit; pre-commit run -a`).  It will either fix issues automatically or give a detailed report of issues.

Regarding this PR, this is really an issue of inconsistent filenames by Jay Anderson.  I spoke with Jay and he has fixed the filename issue with the NIRCam STDPSF files on his website (e.g., https://www.stsci.edu/~jayander/JWST1PASS/LIB/PSFs/STDPSFs/NIRCam/SWC/F070W/).  All of his STDPSF files now have a consistent form of `STDPSF_<detector>_<filter>.fits`.  I also asked him to add header keywords for all his STDPSF files so photutils can avoid having to rely on parsing the filename, which can be fragile as demonstrated here.

The photutils reader should now work for all STDPSF files, so I'm going to close this PR.  Very sorry about that, but thank you again for reporting this issue and submitting a PR!  Please don't hesitate to submit future PRs.  And, please let me know if you are still having any issues.

avapolzin:
Thank you @larrybradley!

PR #2033, opened 2025-04-29
Avoid using `astropy.utils.isiterable()`

eerovaher:
`astropy` is deprecating the function because it duplicates a `numpy` function.


larrybradley (approved):
LGTM, thanks

PR #2039, opened 2025-05-06
Fix not appending to extra_properties when overwrite=True in SourceCatalog

Lyalpha:
When adding a new property to an existing `SourceCatalog` instance with `add_extra_property(..., overwrite=True`), it is not appended to the `_extra_properties` attribute, even if it doesn't already exist, due to: https://github.com/astropy/photutils/blob/0977e52be71ce0c30db8a575ae6ae46efa8f1ab4/photutils/segmentation/catalog.py#L662-L663

If one tries to then overwrite this new property, an unexpected error message occurs as this condition is triggered: https://github.com/astropy/photutils/blob/0977e52be71ce0c30db8a575ae6ae46efa8f1ab4/photutils/segmentation/catalog.py#L628-L630

This update will always append the new property to the `_extra_properties` attribute if it doesn't exist, regardless of `overwrite`, which would be the behaviour at least I was expecting. Apologies if I've misread the intent of this.

This alleviates the use case where one wants to perform an "update or create" action for property that may be called again. Currently this requires checking if the property already exists.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,12 @@ New Features
 Bug Fixes
 ^^^^^^^^^
 
+- ``photutils.segmentation``
+
+  - Fixed an issue where a newly-defined extra property of a
+    ``SourceCatalog`` with ``overwrite=True`` would not be added to
+    the ``extra_properties`` attribute.
+
 - ``photutils.centroids``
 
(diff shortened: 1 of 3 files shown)


larrybradley (approved):
@Lyalpha Many thanks for reporting and fixing this bug!

PR #1288, opened 2022-01-31
Fast build ellipse model

huancho9802:
Hello! This is a pull request in response to #1168 for a faster build_ellipse_model() function, using C and OpenMP. Changes are
- model.py: 
  (1) changed data structure for **result** and **weight** from 2D numpy array to a 1D C-native array
  (2) converted all auxiliary numpy arrays **(intens_array, eps_array, etc)** into C arrays
  (3) ported the for loop into C code
- tests/test_model.py:
  (1) Add new test for a multithreaded version of the function
  (2) Add new test for processing a large array (5000x5000), shows that result is returned very fast
- worker.so: compiled file from C code

Here is the C code for the loop (worker.c):
```C
#include <omp.h>
#include <math.h>
#include <stdlib.h>

#define PI 3.1415926535897932384626433

// MAX FUNCTION
static double max(double a, double b)
{
  if (a > b) return a;
  else return b;
}

// radius function, obtained from code of EllipseGeomtry
// https://photutils.readthedocs.io/en/stable/_modules/photutils/isophote/geometry.html#EllipseGeometry
static double radius(double angle, double sma, double eps)
{
  double val = (sma * (1. - eps) /
                sqrt(pow(((1. - eps) * cos(angle)),2) +
                        pow((sin(angle)),2)));
  return val;
}

// WORKER FUNCTION
void worker(double* result, double* weight, int n_rows, int n_cols, int N, int high_harmonics,
		double* fss_arr, double* intens_arr, double* eps_arr, double* pa_arr,
		double* x0_arr, double* y0_arr, double* a3_arr, double* b3_arr, double* a4_arr, double* b4_arr,
		int nthreads)
{
  int i;
  int n_entries = n_rows * n_cols;

  omp_set_num_threads(nthreads);

  #pragma omp parallel default(shared) private(i)
  {
    // private arrays for each thread to write, calloc auto initializes to 0.0
    double* result_priv = calloc(n_entries, sizeof(double));
    double* weight_priv = calloc(n_entries, sizeof(double));

    #pragma omp for schedule(static) nowait
    for (i = 1; i < N; i++) {
      // variables, obtained at array[index]
      double sma0 = fss_arr[i];
      double intens = intens_arr[i];
      double eps = eps_arr[i];
      double pa = pa_arr[i];
      double x0 = x0_arr[i];
      double y0 = y0_arr[i];

      // Ellipse geometry, _phi_min (defined in original code as 0.05)
[...]

Issue #1168, opened 2021-02-25
build_ellipse_model is very slow


nandinihazra:
Hello Ivo, Larry, Brigitta and others

My advisor Michele Cantiello ( @cantiello) and I have been working on ~3500x3500 pixel FITS images of bright, extended elliptical galaxies in which we fit elliptical isophotes upto maximum semi-major axis lengths of about 500-600 pixels. However, even though the fit itself takes about 20-40 seconds to complete, the build_ellipse_model takes several minutes to run and generate the model image. We spoke to Ivo briefly about this and it seems that the issue might already be on your radar. Since it is posing a significant problem for us at the moment we would really appreciate if you could take a look at it again, or let us know if we can optimise our runs better to get around the bottleneck.

Thank you
Nandini Hazra

larrybradley:
@ibusko

larrybradley:
Ivo has since retired and unfortunately was the author (and expert) for this code.  There is a PR to speed up `build_ellipse model`. Thanks!

PR #2046, opened 2025-05-20
Speed up build_ellipse_model

taranu:
Like #1288  (which seems abandoned), this speeds up build_ellipse_model to address #1168 . However, it is entirely written in Cython and does not require a separate shared library. In local testing on Python 3.12, it's about 130x faster than the current pure Python implementation. I also made the tolerance for one of the unit tests much stricter, because I found that it was passing even when the output was incorrect.

There is a potentially premature optimization to specialize the function depending on whether harmonic arrays are passed or not. The C code generated on my system looked correct, but I didn't actually find any difference in performance between `high_harmonics=False` and True.

I also attempted to parallelize it with Cython's `prange`. However, in my testing on multiple Linux systems, I found it never actually used more than one thread. Additionally, I couldn't figure out how to allow for a `num_threads=None` value in the Cython `build_ellipse_model_c` function, since it's typed as an int. Thus, I'm making this a draft PR to start. If someone more versed in Cython/prange could have a look at that, it would be helpful. Otherwise, the parallel support could simply be dropped.


taranu:
I tested this on two different systems and found that `num_threads` does nothing because it was compiling without `-fopenmp`  and not linking against `libomp` (or `libgomp`). After manually recompiling, parallelism works fine - or at least, runtimes are about ~3.2x faster with 4 threads.

I see that there are [openmp helpers](https://github.com/astropy/extension-helpers/blob/main/extension_helpers/_openmp_helpers.py) in `extension-helpers`, which is a requirement and enabled in pyproject.toml here, but I'm unclear on how to encourage `pip install`  to add the relevant compiler flags. @astrofrog, could you advise?

astrofrog:
Even though extension-helpers includes some OpenMP-related helpers, we haven't really been using them in astropy packages, as we haven't really done much investigation of how this works in terms of building wheels and whether it could cause issues on some platforms. I would personally suggest removing the parallel aspect from this PR for now until we understand better whether we should be using OpenMP or other mechanisms.

taranu:
> I would personally suggest removing the parallel aspect from this PR for now until we understand better whether we should be using OpenMP or other mechanisms.

Thanks for the reply. I was able to compile and install the parallel version via `CFLAGS="${CFLAGS} -fopenmp" LDFLAGS="${LDFLAGS} -lomp" pip install .`, but like you said, I have no idea how to test that across platforms, so I dropped the parallel part from the PR, which should be reviewable now. I'll re-add it on a separate branch if anyone wants to use it in the meantime.

photutils/isophote/ellipse_model.pyx
```diff
@@ -0,0 +1,157 @@
...
+    cnp.ndarray[DTYPE_t, ndim=1] b4_array = None,
+):
+    cdef cython.Py_ssize_t len_sma
+    len_sma = len(finely_spaced_sma)
+    for array in (intens_array, eps_array, pa_array, x0_array, y0_array):
+        if len(array) != len_sma:
+            raise ValueError(f"All input arrays must be same length={len_sma}")
+    harmonics_is_none = (
+        array is None for array in (a3_array, b3_array, a4_array, b4_array)
+    )
```
larrybradley, line 62:
```suggestion
    harmonics_arrays = (a3_array, b3_array, a4_array, b4_array)
    harmonics_is_none = [array is None for array in harmonics_arrays]
```

As a generator, I think the `all(harmonics_is_none)` can short circuit leaving the generator partially consumed, and the `if any(harmonics_is_none)` line will start from the partially consumed generator.  It's better to define `harmonics_is_none` as a list.  This also avoids the generator possibly being calculated twice (once for `all` and once for `any`).

photutils/isophote/ellipse_model.pyx
```diff
@@ -0,0 +1,157 @@
...
+
+    for index in range(1, len_sma):
+        with cython.boundscheck(False):
+            sma = finely_spaced_sma[index]
+            q = 1.0 - eps_array[index]
+            pa = pa_array[index]
+            x0 = x0_array[index]
+            y0 = y0_array[index]
+            intens0 = intens_array[index]
+        phi = 0
```
larrybradley, line 103:
```suggestion
        phi = phi_min
```
I think this should be `phi_min` now?

photutils/isophote/ellipse_model.pyx
```diff
@@ -0,0 +1,157 @@
...
+                    # add up the fractional area contribution to the
+                    # overlapping pixels
+                    weight[j, i] += one_m_fx_t_one_m_fy
+                    weight[j, i + 1] += one_m_fy_t_fx
+                    weight[j + 1, i] += one_m_fx_t_fy
+                    weight[j + 1, i + 1] += fy_t_fx
+
+                # step towards next pixel on ellipse
+                phi = max((phi + 0.75 / r), phi_min)
+                r = sma * q/sqrt((q * cos(phi))**2 + sin(phi)**2)
```
larrybradley, line 149:
```suggestion
                r = sma * q / sqrt((q * cos(phi))**2 + sin(phi)**2)
```

larrybradley (changes requested):
@taranu, many thanks for this PR!  Overall it looks great. I've made a few small comments.  Would you also be able to add a new unit test(s) using non-default `phi_min` and `sma_interval` values?

photutils/isophote/ellipse_model.pyx
```diff
@@ -0,0 +1,157 @@
...
+        intens_func = get_intens_no_harmonics
+        a3 = 0
+        b3 = 0
+        a4 = 0
+        b4 = 0
+        do_harmonics = False
+    else:
+        if any(harmonics_is_none):
+            raise ValueError("Must supply all harmonic arrays if any is not None")
+        for array in (a3_array, b3_array, a4_array, b4_array):
```
larrybradley, line 76:
```suggestion
        for array in harmonic_arrays:
```

photutils/isophote/ellipse_model.pyx
taranu: Actually I read it wrong, the existing code starts at phi=0 and then sets phi to `max((phi + 0.75 / r), phi_min)` for the next step. So in practice it's probably only actually a minimum for the first step, and I'm slightly more confused than before.

photutils/isophote/tests/test_model.py
```diff
@@ -44,11 +44,11 @@ def test_model():
     assert data.shape == model.shape
 
     residual = data - model
-    assert np.mean(residual) <= 5.0
-    assert np.mean(residual) >= -5.0
+    assert np.abs(np.mean(residual)) <= 5.0
 
 
-def test_model_simulated_data():
+@pytest.mark.parametrize('sma_interval', [None, 0.05])
```
larrybradley, line 50:
```suggestion
@pytest.mark.parametrize('sma_interval', [0.1, 0.05])
```

photutils/isophote/tests/test_model.py
```diff
@@ -61,14 +61,19 @@ def test_model_simulated_data():
...
         isophote_list = ellipse.fit_image()
 
+    kwargs = {}
+    if sma_interval is not None:
+        kwargs['sma_interval'] = sma_interval
+
     model = build_ellipse_model(data.shape, isophote_list,
-                                fill=np.mean(data[0:50, 0:50]))
+                                fill=np.mean(data[0:50, 0:50]),
+                                **kwargs)
```
larrybradley, line 70:
```suggestion
                                sma_interval=sma_interval)
```

photutils/isophote/tests/test_model.py
```diff
@@ -61,14 +61,19 @@ def test_model_simulated_data():
         warnings.simplefilter('ignore', RuntimeWarning)
         isophote_list = ellipse.fit_image()
 
+    kwargs = {}
+    if sma_interval is not None:
+        kwargs['sma_interval'] = sma_interval
+
```
larrybradley, line 67: Remove this block and simply use `sma_interval` explicitly.

larrybradley (approved):
@taranu, many thanks for your contribution!

PR #2055, opened 2025-06-13
Re-enable limited api

astrofrog:
This re-enables the limited API now that Cython 3.1.2 is out.

Since last time, we also have Python 3.13 builds on MacOS X and Windows, but @larrybradley feel free to also check locally that everything works fine.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ dev = [
 requires = [
     'setuptools >= 77.0',
     'setuptools_scm >= 8.0',
-    'cython >= 3.1.0,<4',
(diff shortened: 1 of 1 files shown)


larrybradley:
It appears to be due to trying to build `regions` from source (no wheels for 3.13).  Photutils has wheels for 3.13.

astrofrog:
@larrybradley - I'm not quite sure why it suddenly started failing, it looks like somehow GitHub actions might be using a free-threaded version of Python 3.13 on Windows which is very strange. It might be intermittent though. Once the limited API changes are merged for regions we can make a new release which will give us wheels to avoid this issue.

PR #2100, opened 2025-09-13
PSF photometry: Replace compound model with flat model for grouped sources

larrybradley:
With this PR, the PSF photometry classes now use a dynamically generated "flat" model instead of a compound model for grouped sources. This eliminates recursion limits and significantly reduces memory usage for large groups.

Fixes: #1808


larrybradley:
@keflavich and @SterlingYM -- My testing indicates that this PR seems to solve the memory and recursion issues with large source groups.  Can you please also test this with your use cases?

PR #2211, opened 2026-03-04
Initial ASDF extension and tests

nden:
This PR is a contribution to astropy/astropy-project#527 . It implements the initial setup for serializing photutils PSF and aperture objects to ASDF. The proposal mentioned the implementation will be in asdf-astropy. However, after discussion with the ASDF developers we decided to include the implementation in photutils in order to keep the library code in sync with the serialization more easily. This PR adds

- ASDF extension
- Schemas for one PSF model and one aperture
- Converters for the obkects
- Sets up testing of schemas and adds tests for the converters 

I am not able to request reviewers so tagging here @larrybradley  @perrygreenfield @braingram

PR #2268, opened 2026-05-01
Make ASDF extension robust to missing asdf-astropy

larrybradley:
This is a followup to #2211.  Initial ASDF converters were added in #2211, but with `asdf-astropy` as an optional dependency.  Previously, attempting to serialize a `CircularAperture` or `AiryDiskPSF` to ASDF without `asdf-astropy` installed raised an `AsdfSerializationError` with a long traceback.

This PR fixes that by always registering both converters. `CircularApertureConverter` works fully with only `asdf`. `AiryDiskPSFConverter` falls back to a plain `asdf.extension.Converter` base class when `asdf-astropy` is absent, and its to_yaml_tree/from_yaml_tree methods raise a clear `ImportError` pointing the user to `pip install asdf-astropy` instead of a deep traceback.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,7 +24,7 @@ New Features
 ^^^^^^^^^^^^
 
 - Added an ASDF extension to provide converters for photutils aperture
-  and PSF classes. [#2211]
+  and PSF classes. [#2211, #2268]
 
 Bug Fixes
 ^^^^^^^^^
--- a/photutils/converters/__init__.py
+++ b/photutils/converters/__init__.py
@@ -1,20 +1,20 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """
-ASDF converters.
+ASDF converters for photutils objects.
+
+``CircularApertureConverter`` requires only ``asdf``.  ``AiryDiskPSFConverter``
+requires ``asdf-astropy`` for full functionality; it is always registered but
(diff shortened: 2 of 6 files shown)


larrybradley:
Pinging @braingram as a potential reviewer too.

braingram:
This seems like a complicated distinction to make. Is there much benefit in supporting an environment that supports serializing some objects when asdf is installed and not asdf-astropy (which essentially only depends on asdf and astropy)? For example is it possible to construct a `CircularAperture` with quantities? Those wouldn't be supported without asdf-astropy and wouldn't trigger the custom error message added here.

If the goal is to provide a custom message if asdf is used but asdf-astropy is not installed perhaps it would be easier to register a extension that supports all photutils tags and photutils types (that have asdf support) and displays the message on read/write?

larrybradley:
@braingram Many thanks for the review!  `CircularAperture` can't be constructed with Quantities, but `SkyCircularAperture` (which I assume will be coming) requires Quantities (and SkyCoord).

The goal was provide a custom/friendly error message. My solution also doesn't scale -- and @nden is adding more converters (to cover all aperture and psf models).  `sunpy` also has `asdf-astropy` as an *optional* dependency.  I'll take a look to see if they have added any custom error messages or anything similar. I might be overthinking this...

larrybradley:
Sunpy doesn't add any special error handling. I'm going to close this.

PR #2280, opened 2026-05-28
[codex] Fix EllipseSample geometry sma cache reset

### LINKED ISSUE #2105: overwriting `sma` of `EllipseGeometry` in `EllipseSample` constructor
https://github.com/astropy/photutils/blob/31801ba07eb4248149802507569ace6c0ed688b6/photutils/isophote/sample.py#L98-L103
https://github.com/astropy/photutils/blob/31801ba07eb4248149802507569ace6c0ed688b6/photutils/isophote/geometry.py#L150-L154
When the `EllipseSample` constructor is provided with both an `EllipseGeometry` object and an `sma`, it overwrites the `sma` attribute of the `EllipseGeometry` object as expected. However, that is not enough because the `EllipseGeometry` object has already initialized an `initial_polar_radius` attribute using the old `sma`, which will not get updated. Therefore, the first sampled point may have a different radius from the rest. This is not vital when fitting the isophotes, because the samples will be recreated in the iterations; it only causes a noticeable effect in the noniterative mode that simply samples the image without fitting geometric parameters. But, fixing it should help the iterative fitting converge faster.

hgao-astro:
## Summary

Fixes #2105 by resetting EllipseGeometry attributes that are derived from sma when EllipseSample is constructed from an existing geometry and a new semimajor axis.

The root cause was that EllipseSample deep-copied the input geometry and overwrote geometry.sma, but left cached SMA-dependent values such as _area_factor, sector_angular_width, initial_polar_angle, and initial_polar_radius from the original geometry.

## Changes

- Moved SMA-dependent geometry initialization into a private helper on EllipseGeometry.
- Called that helper after EllipseSample overrides the copied geometry sma.
- Added a regression test that compares the copied-and-overridden geometry against a freshly constructed geometry with the target sma.

## Validation

- python -m pytest photutils/isophote/tests/test_sample.py
- python -m pytest photutils/isophote/tests/test_geometry.py photutils/isophote/tests/test_sample.py

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,6 +39,9 @@ Bug Fixes
     an elliptical path that falls entirely (or largely) outside the
     image. [#2283]
 
+  - Fixed ``EllipseSample`` not resetting ``EllipseGeometry`` cached
+    attributes when overriding ``sma``. [#2280]
+
 API Changes
 ^^^^^^^^^^^
 
--- a/photutils/isophote/geometry.py
+++ b/photutils/isophote/geometry.py
@@ -143,10 +143,21 @@ class EllipseGeometry:
         self._phi_min = 0.05
         self._phi_max = 0.2
 
+        self._reset_sma_dependent_attributes()
+
+    def _reset_sma_dependent_attributes(self):
+        """
+        Reset attributes derived from the current semimajor axis length.
+        """
         # variables used in the calculation of the sector angular width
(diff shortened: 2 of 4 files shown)


larrybradley (approved):
Many thanks, @hgao-astro!  I added a changelog entry for this bugfix and will merge when CI passes.

PR #2289, opened 2026-06-08
Gridded PSF performance improvement

keflavich:
This PR attempts to improve performance of GriddedPSF evaluation.

At a low level, some of these steps get ~10x speedups.  However, I think they contribute negligibly to overall speed for most cases, so I'm not sure we should merge this - the slight cost in code complexity may be not worth it.

Evaluating this PR requires tests against a new benchmark:
https://github.com/astropy/photutils-benchmarks/pull/5

Low-level improvement:
```
Test: 4x4 PSF grid, 51-pixel stamps, 2x oversampling, 10,000 evaluate calls

| Operation | Before (us) | After (us) | Speedup |
|-----------|------------|-----------|---------|
| `_find_bounding_points` | 159.52 | 13.11 | **12.17x** |
| `_calc_bilinear_weights + bounding` | 228.72 | 20.75 | **11.02x** |
| `evaluate()` full | 626.39 | 545.66 | **1.15x** |
```

If this 15% improvement really holds up on realistic cases, it's... maybe worthwhile?

Claude Haiku 4.5 contributed a bunch to this.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -56,6 +56,14 @@ New Features
     branch for harmonic corrections. The module is now also marked as
     ``freethreading_compatible=True``. [#2293]
 
+- ``photutils.psf``
+
+  - Improved ``GriddedPSFModel`` evaluation performance by ~15% by
+    replacing the per-call grid lookup with a precomputed grid-cell
+    lookup table and implementing a scalar fast path for the bilinear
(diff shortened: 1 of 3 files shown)


larrybradley (approved):
Thanks, @keflavich.  15% performance increase seems worthwhile.

I added some tests and a changelog entry.

PR #2320, opened 2026-07-13
Fix circular subpixel overlap to be platform independent

larrybradley:
This PR fixes the ``'subpixel'`` overlap method for circular apertures so that it returns identical results across platforms and compilers. Previously, the squared-distance test for each subpixel center could be contracted into a fused multiply-add on platforms with hardware FMA (e.g., Apple silicon), giving slightly different results for subpixel centers lying extremely close to the aperture boundary.

xref: https://github.com/astropy/photutils/actions/runs/29283366986/job/86929792505?pr=2319

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -126,6 +126,16 @@ Bug Fixes
     not input. Previously, an all-NaN error array was returned only
     for sources with no overlap with the data. [#2316]
 
+- ``photutils.geometry``
+
+  - Fixed the ``'subpixel'`` overlap method for circular apertures so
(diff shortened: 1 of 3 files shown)


larrybradley:
Apparently, GCC recombines these statements during optimization (default is `-ffp-contract=fast`), reenabling the FMA.  One possible option is to explicitly set `-ffp-contract=off` when compiling the Cython code.  However, in this case, the doctest example aperture failure is due to having both an integer radius and ycenter, such that the circle edge lands exactly on pixel center(s). I'm closing this PR in favor of simply updating the example in the documentation.

PR #2342, opened 2026-08-04
Fix the multithreshold docstring, which describes a parameter it does not take

karpovantonme:
### Problem

`_SingleSourceDeblender.multithreshold` takes no arguments:

```python
def multithreshold(self):
```

but its docstring documents one:

```
Parameters
----------
deblend_mode : bool, optional
    If `True` then only segmentation images with more than one
    label will be returned. If `False` then all segmentation
    images will be returned.
```

`deblend_mode` appears nowhere else in the package. `grep -rn deblend_mode photutils/` returns this docstring line and nothing more.

The `Returns` section is off too. It promises that only segmentation images with more than one label come back, but the method appends whatever `_detect_sources` returns, without filtering:

```python
segm = _detect_sources(..., relabel=False, return_segmimg=False)
segms.append(segm)
```

and `_detect_sources` documents that with `return_segmimg=False`, "if only one source is found, then `None` is returned". So the list can hold `None` entries, which a caller reading this docstring would not expect.

### Change

Dropped the `Parameters` block, since there are no parameters, and rewrote `Returns` to say that entries can be `None`. Docstring only, no behaviour touched.

### Checks

```
$ pytest --pyargs photutils.segmentation.tests.test_deblend
39 passed

$ ruff check photutils/segmentation/deblend.py
All checks passed!
```

Found with a script that pulls every numpydoc `Parameters` name out of a docstring and compares it against the signature below it. This was the only hit in photutils; `astropy-healpix` came back clean.

### DIFF
--- a/photutils/segmentation/deblend.py
+++ b/photutils/segmentation/deblend.py
@@ -444,19 +444,12 @@ class _SingleSourceDeblender:
 
         This method is useful for debugging and testing.
 
-        Parameters
-        ----------
-        deblend_mode : bool, optional
-            If `True` then only segmentation images with more than one
-            label will be returned. If `False` then all segmentation
-            images will be returned.
-
         Returns
         -------
-        segments : list of 2D `~numpy.ndarray`
-            A list of segmentation images, one for each threshold.
-            Only segmentation images with more than one label will be
-            returned.
+        segments : list of 2D `~numpy.ndarray` or `None`
+            A list of segmentation images, one for each threshold. An
+            entry is `None` where the threshold produced fewer than two
+            sources.
         """
         thresholds = self.compute_thresholds()
         segms = []


larrybradley:
Thank you for looking at the documentation.

Per our project's [Generative AI policy](https://github.com/astropy/astropy-project/blob/main/policies/ai-policy.md), we require contributors to have human ownership over their submissions and be able to personally explain their changes.

Because this edit is inside a non-public, internal function, could you please briefly explain in your own words what specific task or feature you were working on in our codebase that led you to discover this unused parameter docstring?

We will hold off on reviewing this change until we receive your response. Thank you!

karpovantonme:
Fair question, and the honest answer is that I was not working in your codebase at all 🙃

I wrote a checker that compares numpydoc `Parameters` against the actual signature, ran it across a set of projects, and photutils came up in the results. I opened the file, saw a parameter documented on a function that does not take it, and sent the fix. That is the whole story: no task, no feature, just the checker plus me reading the hit.

If that does not meet the bar in your policy, I understand, and I will leave astropy alone from here. No hard feelings either way, and thanks for taking the time to explain the policy rather than just closing it

larrybradley:
Thanks. To be direct: are you a human, or is this an automated account/AI-generated response? If you are a human contributor, please reply with a single sentence containing only the name of the file you edited and the exact line numbers you changed. Please do not include any other conversational text or pleasantries.

larrybradley:
I'm closing this PR as AI-generated without attribution (also see https://github.com/astropy/astroquery/pull/3641#issuecomment-5187454470).

